PR #115, opened 2014-04-25
Task export

adamruzicka:
- [x] Support for exporting delayed tasks
- [x] Code cleanup
- [x] Fix tests
- [x] HTML export


adamruzicka:
Rebased on latest-greatest, added exapmle api endpoint, pretty much WIP again

lib/dynflow/web/console.rb
```diff
@@ -49,6 +50,21 @@ class Console < Sinatra::Base
         erb :worlds
       end
 
+      get('/export') do
+        set_download_headers('index.tar.gz')
+        Exporters::Tar.full_html_export(find_execution_plans(true, false), self)
```
iNecas: I'm not sure we should set `show_all` to true and `paginate` to false. This could pretty much get down the whole instance with rich execution plans history.

test/exporter_test.rb
```diff
@@ -0,0 +1,100 @@
...
+          exported_plan = @exporter.export
+          finalize = exported_plan[:phase][:finalize]
+          finalize[:type].must_equal 'sequence'
+          finalize[:steps].count.must_equal execution_plan.finalize_flow.flows.count
+        end
+
+      end
+    end
+
+    describe ::Dynflow::Exporters::Tar do
```
iNecas: Do you plan on writing JSON and HTML exporter tests as well? I would be ok with some more high level tests, with more end-2-end tests rather than the stubbing once, unless the time penalty was too high.

adamruzicka: Added some more tests for the JSON and HTML exporters

lib/dynflow/web/console.rb
adamruzicka: Those were meant as an example how to use the tar exporter, maybe removing the index-like export api endpoints and having a separate PR for a well-thought api with filtering would be better

iNecas: +1 from removing from here then

iNecas:
Summary of what discussed offline: rather than an end-user script, it seem better would be to provide a rake task constructor (similar to http://rake.rubyforge.org/classes/Rake/TestTask.html) that would be possible to use from the project it want's to leverage it, and it would get the dynflow world instance to be used from there

lib/dynflow/exporters/tar.rb
```diff
@@ -0,0 +1,81 @@
+require 'zlib'
+require 'archive/tar/minitar'
```
iNecas: Would we try to use http://ruby-doc.org/stdlib-2.2.2/libdoc/rubygems/rdoc/Gem/Package/TarWriter.html, to minimize external dependencies?

lib/dynflow/utilities/export.rb
```diff
@@ -0,0 +1,90 @@
+module Dynflow
+  module Utilities
+    class Export
```
iNecas: I guess this could turn into the rake task class

bin/dynflow-export
```diff
@@ -0,0 +1,57 @@
...
+utils = ::Dynflow::Utilities::Export.new
+
+OptionParser.new do |opts|
+  opts.banner = 'Usage: dynflow-export [options]'
+
+  opts.on('-h', '--help', 'Prints help') do
+    puts opts
+    puts "\nAll options can be subtituted by environment variables, however CLI arguments take precedence."
+    puts <<-DESC
+    * TASK_SEARCH     : scoped search filter (example: 'id = "d99bb632-2414-4c2b-af9a-af475b2a4723"')
```
iNecas: it's not scoped search, right? What's the expected syntax of search, when not in tasks?

lib/dynflow/exporters/abstract.rb
```diff
@@ -0,0 +1,95 @@
...
+          :plan   => execution_plan,
+          :result => nil
+        }
+        self
+      end
+
+      # Add execution plan to export by id
+      def add_id(execution_plan_id)
+        @index[execution_plan_id] = {
+          :plan   => nil,
```
iNecas: please explain in the comment, where the `:plan` will get filled in?, and the whole difference between `add` and `add_id`

lib/dynflow/exporters/abstract.rb
```diff
@@ -0,0 +1,104 @@
...
+
+      # The same as #add, but takes Array[ExecutionPlan]
+      def add_many(execution_plans)
+        execution_plans.each { |plan| add plan }
+        self
+      end
+
+      # The same as #add_id, but takes Array[String]
+      def add_many_ids(execution_plan_ids)
+        execution_plan_ids.each { |plan_id| add_id plan_id }
```
lzap: Can you measure if http://api.rubyonrails.org/classes/ActiveRecord/Batches.html helps here (and above) a bit? It loads records in batches rather that one by one from database.

lib/dynflow/tasks/export_task.rb
```diff
@@ -0,0 +1,146 @@
...
+                  when 'json'
+                    ::Dynflow::Exporters::Tar.full_json_export plans
+                  when 'csv'
+                    ::Dynflow::Exporters::CSV.new
+                      .add_many(plans).result
+                  else
+                    raise "Unknown export format '#{format}'"
+                  end
+        File.write(task_file, content)
+        puts "Exported tasks as #{task_file}"
```
lzap: This I believe puts all records into memory and then flushes it to disk. Why you haven't designed this as a stream of data? That would be much more robust and memory efficient. Task export on an instance can easily generate 200 MB of compressed (`xz`) data, that is gigabyte of HTML. This can lead to swapping and possibly very slow export on a production system under load.

I suggest a change to read records one by one streaming them to disk, otherwise we are not getting anywhere with this.

lib/dynflow/exporters/abstract.rb
adamruzicka: This place doesn't interact with the database at all, everything thouching the db happens [here](https://github.com/Dynflow/dynflow/pull/115/files#diff-4ca04e4243569cbfd52809ce04bd1743R97). Dynflow uses Sequel (instead of ActiveRecord) to work with the database and in this case the retrievals happen in batches of 1000 records afaik.

Introducing AR would be quite big change

lib/dynflow/tasks/export_task.rb
adamruzicka: It indeed puts everything into memory and then dumps it. The main concern of this PR was to move the Dynflow specific bits and pieces from Foreman Tasks, so the same logic could be used for exporting tasks from Foreman Tasks as well as Smart Proxy, while being easily extensible.

But agreed, performance here isn't exactly great, so I'll see what can be done about it.

lib/dynflow/exporters/abstract.rb
lzap: Ah right :-) Thanks.

lib/dynflow/tasks/export_task.rb
lzap: Streaming would be better, I do believe this patch is very useful and will improve tasks export a lot! Please keep going.

lib/dynflow/tasks/export_task.rb
```diff
@@ -65,32 +70,50 @@ def task_ids
...
-                  when 'csv'
-                    ::Dynflow::Exporters::CSV.new
-                      .add_many(plans).result
-                  else
-                    raise "Unknown export format '#{format}'"
-                  end
-        File.write(task_file, content)
+        puts "Exporting #{plans_or_ids.count} tasks"
+        File.open(task_file, 'w') do |f|
+          f = Zlib::GzipWriter.new(f) if compress?
```
iNecas: refactoring suggestiong: extract the potential comprssion into `with_comprission(f, block&)` that would look something like:

```ruby
def with_compression(f, &block)
  if compress?
    begin
      gzip_f = Zlib::GzipWriter.new(f)
      yield gzip_f
    ensure
      gzip_f.close
    end
  else
    yield f
  end
end
```

lib/dynflow/exporters/abstract.rb
```diff
@@ -2,103 +2,41 @@ module Dynflow
...
+
+      # Used when streaming collection, should be a 3-tuple containing
+      #   what is printed before items
+      #   what is printed between the items
+      #   what is printed after the items
+      def brackets
+        []
       end
 
+      # def self.export_execution_plan_id(world, execution_plan_id, options = {})
```
iNecas: Should this code be uncommented or removed?

adamruzicka: Removed, exporting one single task is much simpler now so it doesn't need a helper method

lib/dynflow/tasks/export_task.rb
adamruzicka: Good point, looks much cleaner this way

lzap:
Re streaming - it was implemented here FYI (not difficult): https://github.com/theforeman/foreman/pull/4270

iNecas:
Related topic https://bugzilla.redhat.com/show_bug.cgi?id=1439230

pitr-ch:
Just saying "Hi" 👋 Quite a surprise getting a notification after 8 years 🤣 Have a good week!

PR #128, opened 2014-09-04
Allow deduplication of actions in run and finalize phases

adamruzicka:


ezr-ondrej:
@adamruzicka do we still want to get this in?

adamruzicka:
Not really, it was a fun exercise but would probably do more harm than good in the long run. Closing.

PR #135, opened 2015-01-23
Better handling of databse connections errors

iNecas:
1. Retry several times when db error occurs
2. Terminate when the retries don't help
3. Let the executor to finish just after the workers are done:
   don't wait for the whole task to finish.

### DIFF
--- a/lib/dynflow/errors.rb
+++ b/lib/dynflow/errors.rb
@@ -24,5 +24,13 @@ module Dynflow
         "#{self.class.inspect}: #{message}"
       end
     end
+
+    class PersistenceError < StandardError
+      def self.delegate(original_exception)
+        self.new("caused by #{original_exception.class}: #{original_exception.message}").tap do |e|
+          e.set_backtrace original_exception.backtrace
+        end
+      end
+    end
   end
 end
--- a/lib/dynflow/world.rb
+++ b/lib/dynflow/world.rb
@@ -10,6 +10,7 @@ module Dynflow
       @logger_adapter      = Type! option_val(:logger_adapter), LoggerAdapters::Abstract
       @transaction_adapter = Type! option_val(:transaction_adapter), TransactionAdapters::Abstract
       persistence_adapter  = Type! option_val(:persistence_adapter), PersistenceAdapters::Abstract
+      persistence_adapter.logger ||= logger
       @persistence         = Persistence.new(self, persistence_adapter)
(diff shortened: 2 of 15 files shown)


lib/dynflow/errors.rb
```diff
@@ -24,5 +24,13 @@ def inspect
...
       end
     end
+
+    class PersistenceError < StandardError
+      def self.delegate(original_exception)
+        self.new("caused by #{original_exception.class}: #{original_exception.message}").tap do |e|
+          e.set_backtrace original_exception.backtrace
+        end
+      end
+    end
```
pitr-ch: what about `justified` gem?

lib/dynflow/world.rb
```diff
@@ -32,7 +33,7 @@ def default_options
     end
 
     def clock
-      @clock ||= Clock.new(logger)
+      @clock_barrier.synchronize { @clock ||= Clock.new(logger) }
```
pitr-ch: Could we set up the clock in the `initialize` method? Then we can avoid the barrier.

lib/dynflow/errors.rb
iNecas: Adding additional dependency is not justified for this simple code

lib/dynflow/world.rb
iNecas: I was thinking about it, but in the remote executor case, the clock is seldom used: so we will have additional thread in every client instance.

iNecas:
Resolved issues with event futures not getting resolved occasionally on termination

lib/dynflow/errors.rb
pitr-ch: ok :D

lib/dynflow/world.rb
pitr-ch: hm its sleeping thread vs synchronizing all the time. Probably not important.

pitr-ch:
ack

iNecas:
I've two more commits for exiting the process when terminating and handling an edge case of some plan being already in paused state when finishing the termination

pitr-ch:
:+1:

iNecas:
@pitr-ch one last update: I added the termination when the db connection fails constantly only when happens inside the executor: the client part doens't necessary need to fail when this occurs.

pitr-ch:
LGTM

PR #136, opened 2015-01-31
Fix tests on random unresolved future

iNecas:
There was a case where an event came right at the time of the core being
terminated, and ended up in the unresolved state.

Handling this in actor by dead letter routing is not as easy as it seems (as
making sure to read all the messages after termination turns out to not be so
straight forward). So for now we just make sure to refuse all the incoming
events as soon as possible.

### DIFF
--- a/lib/dynflow/world.rb
+++ b/lib/dynflow/world.rb
@@ -107,7 +107,14 @@ module Dynflow
     end
 
     def event(execution_plan_id, step_id, event, future = Future.new)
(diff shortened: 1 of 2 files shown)


iNecas:
After a year, there should be no more accidental unresolved futures at the end of the tests. Looking forward to debugging the same with the new executor:)

pitr-ch:
:+1: :+1: :+1:

PR #140, opened 2015-02-24
Polling action tweaks

adamruzicka:
Polling action:
- made ```done?``` method able to raise an exception and stop the action
- added ```on_finish``` method


iNecas:
Nice!

PR #145, opened 2015-03-13
Sub plans support

iNecas:
This patch adds support for triggering execution plans from a run phase
(a.k.a sub-plans) and waiting for the sub-plans to finish.

It's as simple as:

```ruby
   class ActionWithSubPlan < Dynflow::Action
     include Dynflow::Action::WithSubPlans

      def create_sub_plans
       10.times.map { trigger(SomeAction, input) }
     end
   end
```

The `Action::WithSubPlans` defines the `run` method and expects the action
to define `create_sub_plans` to return one or many triggered actions. The
`WithSubPlans` module takes care of waiting for the sub plans to finish,
and succeeding or failing after that.

The resuming behaviour can be also customized, by overriding the `resume`
method. By default, it re-creates the sub plans in case none of the sub
plans was planned successfully. Otherwise it checks, that all the sub plans
have already finished and fails when they haven't.

To support this, the caller_execution_plan_id and caller_action_id
attributes were added to the action, to track this relation. The relation
between actions in one execution plan is tracked the same way as well, with
the prospect of using this data instead of the data in execution plan later
(but we are not there yet).

In the web console, the Action::WithSubPlans shows link to the list of sub
plans in the console.

Also, an example of the sub plans is available in examples directory for
seeing this functionality in cation.

### DIFF
--- a/lib/dynflow/action/with_sub_plans.rb
+++ b/lib/dynflow/action/with_sub_plans.rb
@@ -6,17 +6,18 @@ module Dynflow
     end
 
     def run(event = nil)
-      case(event)
-      when nil
-        if output[:total_count]
-          resume
-        else
-          initiate
-        end
-      when SubPlanFinished
-        mark_as_done(event.execution_plan_id, event.success)
-        try_to_finish or suspend
-      end
+      match event,
+            (on nil do
+               if output[:total_count]
+                 resume
+               else
+                 initiate
+               end
+             end),
+            (on SubPlanFinished do
+              mark_as_done(event.execution_plan_id, event.success)
+              try_to_finish or suspend
+             end)
     end
 
     def initiate
@@ -25,7 +26,7 @@ module Dynflow
       wait_for_sub_plans(sub_plans)
     end
 
-    # @api override when the logic for the initiation of the subtasks
+    # @abstract when the logic for the initiation of the subtasks
     #      is different from the default one.
     # @returns a triggered task or array of triggered tasks
     # @example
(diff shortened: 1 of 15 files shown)


lib/dynflow/action/with_sub_plans.rb
```diff
@@ -0,0 +1,161 @@
...
+      end
+    end
+
+    def initiate
+      sub_plans = create_sub_plans
+      sub_plans = Array[sub_plans] unless sub_plans.is_a? Array
+      wait_for_sub_plans(sub_plans)
+    end
+
+    # @api override when the logic for the initiation of the subtasks
```
pitr-ch: I think this should be `@abstract` tag.

lib/dynflow/action/with_sub_plans.rb
```diff
@@ -0,0 +1,161 @@
...
+      when nil
+        if output[:total_count]
+          resume
+        else
+          initiate
+        end
+      when SubPlanFinished
+        mark_as_done(event.execution_plan_id, event.success)
+        try_to_finish or suspend
+      end
```
pitr-ch: missing else branch, :wink: `match`

lib/dynflow/action/with_sub_plans.rb
```diff
@@ -0,0 +1,161 @@
...
+    end
+
+    def wait_for_sub_plans(sub_plans)
+      output.update(total_count: 0,
+                    failed_count: 0,
+                    success_count: 0)
+
+      planned, failed = sub_plans.partition(&:planned?)
+
+      sub_plan_ids = ((planned + failed).map(&:execution_plan_id))
```
pitr-ch: extra ()

pitr-ch: @iNecas doing lisp much lately? :)

iNecas:
PR comments addressed

pitr-ch:
Thanks! ACK

PR #147, opened 2015-03-18
Attempt to recreate socket when unavailable

eLobato:
I wanted to write a test but I'm not entirely sure how to proceed, help appreciated.

### DIFF
--- a/lib/dynflow/executors/remote_via_socket/core.rb
+++ b/lib/dynflow/executors/remote_via_socket/core.rb
@@ -146,6 +146,10 @@ module Dynflow
         rescue SystemCallError, IOError => error
           logger.warn error
           false
+        rescue Errno::ENOENT, Errno::ECONNREFUSED => error # No such file or directory/ Conn. refused
+          logger.warn 'Socket unavailable: Attempting to recreate the socket for the next connection'
+          Dynflow::Listeners::Socket.new(@world, @socket_path)
+          false
         rescue => error
           logger.fatal error
           raise error


iNecas:
What exactly is the case you're trying to solve? The socket should be owned and created by the server (the remote executor is here the client-side of the whole stack) It's also going to be replaced by the database-based communication in next major version.

eLobato:
I wasn't exactly sure how to solve the problem. Somehow I started to get
Connection refused - "/usr/share/foreman/tmp/sockets/dynflow_socket"
on every Dynflow connection and it wasn't fixed until I restarted
foreman-tasks that the problem got fixed. I don't know yet what caused
that but it's impossible to know how to recreate the socket.

I thought first about logging it better and this was the culprit in the
backtrace so I thought maybe creating the socket here would work, any
ideas?
On 03/18, Ivan Necas wrote:
> What exactly is the case you're trying to solve? The socket should be owned and created by the server (the remote executor is here the client-side of the whole stack) It's also going to be replaced by the database-based communication in next major version.
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/Dynflow/dynflow/pull/147#issuecomment-82957918

--
Daniel Lobato Garcia

@eLobatoss
blog.daniellobato.me
daniellobato.me

GPG: http://keys.gnupg.net/pks/lookup?op=get&search=0x7A92D6DD38D6DE30
Keybase: https://keybase.io/elobato

iNecas:
Try looking into /usr/share/foreman/tmp/pids/dynflow_executor.output for more hints (we're tracking a BZ to get that file into better placement https://bugzilla.redhat.com/show_bug.cgi?id=1194607)

It seems the executor is getting terminated for some reason: we need to figure out why. I don't think this PR would help with the issue you're seeing

eLobato:
Ok, feel free to close. Maybe log something at that point related to how
to recreate the socket?

On 03/18, Ivan Necas wrote:
> Try looking into /usr/share/foreman/tmp/pids/dynflow_executor.output for more hints (we're tracking a BZ to get that file into better placement https://bugzilla.redhat.com/show_bug.cgi?id=1194607)
>
> It seems the executor is getting terminated for some reason: we need to figure out why. I don't think this PR would help with the issue you're seeing
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/Dynflow/dynflow/pull/147#issuecomment-83021209

--
Daniel Lobato Garcia

@eLobatoss
blog.daniellobato.me
daniellobato.me

GPG: http://keys.gnupg.net/pks/lookup?op=get&search=0x7A92D6DD38D6DE30
Keybase: https://keybase.io/elobato

iNecas:
As I said, we are slowly getting away from the socket-based communication, we will document properly the new approach for remote/mutli - executors setup

PR #139, opened 2015-02-23
Multiple executors support and migration to concurrent ruby

iNecas:
TODO:

- [x] add db matrix to travis (postgres differs from sqlite now)
- [x] use Concurrent::Event for PostgresListener @started variable
- [x] extract ping from world to connector
- [x] dont send envelopes to world
- [x] resolve occasional test failure
- [x] mark coordinator and connector sperific sequel adapter features as optional
- [x] update documentation
- [x] update foreman-tasks to use this and more real-world testing
- [x] resolve the alternative to `observer` and `dataflow` (@pitr-ch)

PR #148, opened 2015-04-05
Pages in progress

iNecas:
This are proposed changes to the docs as it will look like when https://github.com/Dynflow/dynflow/pull/139 gets merged


iNecas:
The changes got in together wiht #139

PR #149, opened 2015-04-05
More info about World

ares:
Meanwhile you covered what I wanted to push :-) So at least one sentence I had in my version that might or might not be useful.


iNecas:
Thanks

PR #150, opened 2015-05-06
Allow using suspended_action reference before calling the suspend method

iNecas:
The usage of suspend with block to get the suspended action reference is a bit
clumsy. This should help a bit with the readability.


pitr-ch:
:+1:

PR #151, opened 2015-05-12
adding simple polling timeout mechanism

jlsherrill:
This adds the ability for an action to call schedule_timeout(50), to
schedule when the polling should stop and the task be marked as failed.
In this simple implementation the above would cause the task to fail after 50
seconds.  Actions could override process_timeout to do more advanced things

### DIFF
--- a/lib/dynflow/action/polling.rb
+++ b/lib/dynflow/action/polling.rb
@@ -1,5 +1,11 @@
+require 'dynflow/action/timeouts'
+
 module Dynflow
   module Action::Polling
+
+    def self.included(base)
+      base.send :include, Action::Timeouts
+    end
 
     Poll = Algebrick.atom
 
@@ -13,6 +19,9 @@ module Dynflow
         end
       when Poll
         poll_external_task_with_rescue
+      when Action::Timeouts::Timeout
+        process_timeout
+        suspend
       else
         raise "unrecognized event #{event}"
       end
--- a/lib/dynflow/action/timeouts.rb
+++ b/lib/dynflow/action/timeouts.rb
@@ -1,0 +1,13 @@
+module Dynflow
+  module Action::Timeouts
+    Timeout = Algebrick.atom
+
+    def process_timeout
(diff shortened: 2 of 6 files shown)


jlsherrill:
example of a more advanced implementation:  https://github.com/Katello/katello/pull/5225

this uses process_timeout to do a 2-phase timeout

lib/dynflow/action/polling.rb
```diff
@@ -13,6 +14,9 @@ def run(event = nil)
         end
       when Poll
         poll_external_task_with_rescue
+      when Timeout
+        process_timeout
+        poll_external_task_with_rescue
```
iNecas: This line is unreachable, rigth (since the process_timeout will fail)

iNecas:
Looks good. One suggestion: extracting the timeout-specific methods into `Action::Timeout` and including that into the Polling, so that other actions can use that as well: might be useful not just for polling

lib/dynflow/action/polling.rb
jlsherrill: Only if process timeout throws an exception (in the default case that is true, but in the 'advanced' cases where process_timeout is overridden, maybe not).

iNecas:
Would you mind adding some tests into https://github.com/Dynflow/dynflow/blob/master/test/action_test.rb, you can inspire by https://github.com/Dynflow/dynflow/blob/master/test/action_test.rb#L256: simply defining a dummy polling action that would define the timeouts and would simulate the hanging, and then checking that it got paused after the timeout (the timeout might be really short so that the tests are still fast)

iNecas:
The tests would then be also a good example on how to use this feature

lib/dynflow/action/polling.rb
iNecas: shouldn't that then be part of the overridden process_timeout the responsibility to poll?

iNecas:
Thanks @jlsherrill

PR #152, opened 2015-06-16
fix spelling of assert methods

sdherr:
The names of the some of the assert methods in assertions.rb are misspelled. This pull request fixes the spelling while also aliasing the old names so that you won't break anyone who is currently using them.


iNecas:
Thanks. Would you mind creating temporary methods for the old variant, with deprecated warning, just to let people to use the new version without getting errors immediately: we can remove that the next minor version.

iNecas:
Nm: i've just noticed you've done so :)

sdherr:
I think you should just leave the aliases. There's no reason to remove the
old names and force dependant code to be updated, even in the next major
version.

PR #153, opened 2015-06-18
Allow describing the action state in more user-oriented way

iNecas:
The 'suspended' state might be misleading for some users not familiar much
with the dynflow terminology.


thomasmckay:
ack w/ other PRs

PR #157, opened 2015-07-21
Refs #10753 - Allows setting future time of execution

adamruzicka:
- [x] basic implementation
- [x] tests
- [x] docs


doc/pages/source/documentation/index.md
```diff
@@ -268,6 +270,38 @@ def self.trigger_task(async, action, *args, &block)
 end
 ```
 
+#### Scheduling
+
+Scheduling an action means setting it up to be triggered at set time in future.
+Any action can be scheduled by calling:
+
+```ruby
+world_instance.schedule(AnAction, start_at, start_before, *args)
```
iNecas: Two things to suggest:
* using the scheduling options as hash: if we introduce another option, we don't break the backward compatibility with the users' code
* Use specific values instead of the variables in the example, so that the users know imediately, whether Time object or String in some format is expected

 `world_instance.schedule(AnAction, {start_at: Time.now + 360, start_before: Time.now + 400}, *args)`

doc/pages/source/documentation/index.md
```diff
@@ -268,6 +270,38 @@ def self.trigger_task(async, action, *args, &block)
...
+```
+
+This snippet of code would schedule `AnAction` with arguments `args` to be executed
+in the time interval between `start_at` and `start_before`. Setting `start_before` to `nil`
+would schedule this action without the upper time limit.
```
iNecas: Is it `upper` or `bottom`, maybe timeout limit is better describing what it does?

doc/pages/source/documentation/index.md
```diff
@@ -268,6 +270,38 @@ def self.trigger_task(async, action, *args, &block)
...
+
+When an action is scheduled, minimal execution plan is created for it and a schedule record
```
iNecas: What about starting describing first how it behaves from the users perspective:
"When an action is scheduled, execution plan object is created (state = scheduled), but it doesn't run the plan phase yet: the planning happens when the `start_at` time comes."
Followed by what happens when the scheduler is down. Also, I would prefer marking the plan as stopped/error, instead of deleting it:
"When the planning doesn't happened before the `start_before` time, the execution plan is marked as failed (state = stopped, result = error)"

Then we could talk about the serialization and we would talk about the Scheduler itself just at the end, as this is more an implementation detail than something the user is really interested in.

examples/future_execution.rb
```diff
@@ -0,0 +1,73 @@
...
+end
+
+class CustomPassedObjectSerializer < ::Dynflow::Serializers::Abstract
+  def serialize
+    object = @args.first
+    # Serialized output can be anything that is representable as JSON: Array, Hash...
+    { :id => object.id, :name => object.name }
+  end
+
+  def deserialize(args)
```
iNecas: `s/args/serialized_args/`

doc/pages/source/documentation/index.md
adamruzicka: Makes sense, will do

lib/dynflow/config.rb
```diff
@@ -85,6 +85,14 @@ def method_missing(name)
       true
     end
 
+    config_attr :scheduler_options, Hash do
+      { :poll_interval => 15, :time_source => -> { Time.now.utc } }
+    end
+
+    config_attr :scheduler_class, Class, NilClass do
+      Schedulers::Polling
```
iNecas: why not having just one config option: `:scheduler` and expecting it to provide the scheduler object itself?

lib/dynflow/persistence_adapters/sequel.rb
```diff
@@ -31,7 +31,8 @@ def ordering_by
                     action:              %w(caller_execution_plan_id caller_action_id),
                     step:                %w(state started_at ended_at real_time execution_time action_id progress_done progress_weight),
                     envelope:            %w(receiver_id),
-                    coordinator_record:  %w(id owner_id class) }
+                    coordinator_record:  %w(id owner_id class),
+                    scheduled:           %w(execution_plan_uuid start_at start_before args_serializer)}
```
iNecas: `s/args_serializer/serialized_args`

iNecas: nm, I've probably just misunderstood the content of this param

lib/dynflow/scheduled_plan.rb
```diff
@@ -0,0 +1,59 @@
...
+       recursive_to_hash :execution_plan_uuid => @execution_plan_uuid,
+                         :start_at            => time_to_str(@start_at),
+                         :start_before        => time_to_str(@start_before),
+                         :args                => @args_serializer.serialize,
+                         :args_serializer     => @args_serializer.class.name
+    end
+
+    # @api private
+    def self.new_from_hash(world, hash, *args)
+      serializer = hash[:args_serializer].constantize.new
```
iNecas: What happens, when the serializer class is removed in the mean time: think of a case, when somebody would rename the serializer in new version of the software, and one updates the application, while something scheduled for next week: Just makring the plan as failed would work here.

lib/dynflow/serializers/abstract.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      def initialize(*args)
+        @args = args
+      end
+
+      def serialize
+        raise NotImplementedError
+      end
+
+      def deserialize(args)
```
iNecas: `s/args/serialized_args`

iNecas:
Would you mind extending the middleware to support the `schedule` phase?

doc/pages/source/documentation/index.md
adamruzicka: Looks like I will have to reword it completely. 

The execution plan is in fact marked as failed (stopped+error) on timeout, the only thing that gets removed is the schedule record.

examples/future_execution.rb
adamruzicka: Why? When the serializer is created it gets the arguments it should pass to the action.  It stores them as they are, they are not serialized yet. The serialization itself happens in this method, using the 'raw' arguments.

lib/dynflow/scheduled_plan.rb
adamruzicka: I haven't though of this, your suggestion seems like a reasonable solution to it.

doc/pages/source/documentation/index.md
iNecas: No need to mention the schedule record that much. Maybe just the fact, that the 'schedule record' is there only at the time the schedule is active, might be enough for the user to know

adamruzicka:
Support for schedule in middleware in next commit

doc/pages/source/documentation/index.md
```diff
@@ -851,12 +885,12 @@ Each **Action phase** can be in one of the following states:
 
 ### Error handling
 
-If there is an error risen in **`plan` phase**, the error is persisted in the Action object 
-for later inspection and it bubbles up in `World#trigger` method which was used to trigger 
-the action leading to this error. 
+If an error rises in **`plan` phase**, it is persisted in the Action object
```
iNecas: `error raises`?

lib/dynflow/schedulers/abstract.rb
```diff
@@ -0,0 +1,17 @@
+module Dynflow
+  module Schedulers
+    class Abstract
+
+      attr_reader :core
+
+      def ask(*args)
```
iNecas: What about making the interface more explicit about what features/methods to implement and less about the actor being used at the backend? If I implemented new interface, I would welcome to see the Abstract class and see, what methods I'm expected to implement, no matter what interface I choose.

iNecas:
Thanks for addressing the comments, some more added (but that should be it:), also, it would be valuable to see the scheduling attributes in the dynflow console

lib/dynflow/world.rb
```diff
@@ -25,6 +25,7 @@ def initialize(config)
       @connector            = config_for_world.connector
       @middleware           = Middleware::World.new
       @client_dispatcher    = spawn_and_wait(Dispatcher::ClientDispatcher, "client-dispatcher", self)
+      @scheduler            = try_spawn_scheduler(config_for_world)
       @meta                 = config_for_world.meta
```
iNecas: Adding

```ruby
@meta['scheduler']    = true if @scheduler
```

would help investigating what executor is handling the sechduling

lib/dynflow/world.rb
```diff
@@ -25,6 +25,7 @@ def initialize(config)
       @connector            = config_for_world.connector
       @middleware           = Middleware::World.new
       @client_dispatcher    = spawn_and_wait(Dispatcher::ClientDispatcher, "client-dispatcher", self)
+      @scheduler            = try_spawn_scheduler(config_for_world)
```
iNecas: We should limit the scheduler to run only on words with executor: the reason: the worlds without executor are usualy the client once, that are not expected to run long-term.

lib/dynflow/world.rb
adamruzicka: Added

lib/dynflow/world.rb
adamruzicka: Makes sense, added this check to the `try_spawn_scheduler` method

iNecas:
All comments addressed, please squash before we merge

iNecas:
Thanks you @adamruzicka , merging now

PR #159, opened 2015-07-28
Fix #158 - convert step inputs column to mediumtext on MySQL/MariaDB

### LINKED ISSUE #158: Support for huge inputs with mysql
Seems like dynflow_actions.data column is a TEXT type which on MySQL limits the input to 64kB. For large inputs (e.g. foreman_chef attributes import) it's not enough. I'd suggest using MEDIUMTEXT, 16MB *ought to be enough for anybody*

ares:


lib/dynflow/persistence_adapters/sequel_migrations/006_fix_data_length.rb
```diff
@@ -0,0 +1,15 @@
+Sequel.migration do
+  up do
+    alter_table(:dynflow_steps) do
+      if Sequel::Model.db.database_type == :mysql
```
iNecas: using `@db` might be a bit safer here, as I'm not sure how it is with the global attribute `Sequel::Model.db` when there would be more sequel instances connected to different adapters.

ares: How do I set this `@db`? It does not seem to be present in migration block. I suppose I should ask world instance but I'm not sure how to get it in migration.

adamruzicka:
`@db` should be accessible and already set to the db being migrated inside the `alter_table` blocks

lib/dynflow/persistence_adapters/sequel_migrations/006_fix_data_length.rb
adamruzicka: @db should be accessible and already set to the db being migrated inside the alter_table blocks

lib/dynflow/persistence_adapters/sequel_migrations/006_fix_data_length.rb
```diff
@@ -0,0 +1,15 @@
...
+        set_column_type :data, :mediumtext
+      end
+    end
+  end
+
+  down do
+    if Sequel::Model.db.database_type == :mysql
+      set_column_type :data, :text
```
adamruzicka: Wouldn't this need to be wrapped in an `alter_table` block or given a table symbol directly?

ares:
ok fixed both, thanks

iNecas:
Thanks @ares will merge after travis finishes

lib/dynflow/persistence_adapters/sequel_migrations/006_fix_data_length.rb
```diff
@@ -0,0 +1,17 @@
...
+  up do
+    alter_table(:dynflow_steps) do
+      if @db.database_type == :mysql
+        set_column_type :data, :mediumtext
+      end
+    end
+  end
+
+  down do
+    alter_table(:dynflow_steps)
```
adamruzicka: missing `do` here

ares: Gah, that happens when you're in hurry. Sorry will fix it tomorrow

ares:
[test]

iNecas:
Thank you @ares

PR #161, opened 2015-08-10
Better progress reporting for pending/scheduled plans

iNecas:
They should be always 0.

Also, I hit an issue with a state, where the plan was in scheduled state, but
the scheduler record was not there: I updated the dynflow console to handle
this error better.

### DIFF
--- a/test/future_execution_test.rb
+++ b/test/future_execution_test.rb
@@ -15,7 +15,7 @@ module Dynflow
         end
 
         let(:world) { WorldFactory.create_world }
-        let(:plan) do
+        let(:scheduled_plan) do
           scheduled = world.schedule(::Support::DummyExample::Dummy, { :start_at => @start_at })
           scheduled.must_be :scheduled?
           world.persistence.load_scheduled_plan(scheduled.execution_plan_id)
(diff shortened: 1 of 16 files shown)


adamruzicka:
Looks good to me.

Do you know how to reproduce the issue with missing schedule record?

iNecas:
@adamruzicka after writting tests I've realised we should really handle the errors in the serializer better, which lead me to some bigger changes, including the serializer interface, so that we can perform the serialization while catching the errors, and then access the serialized args when creating the schedule object. I've also hit some other issues, that I've fixed as well

test/future_execution_test.rb
```diff
@@ -94,13 +105,13 @@ module FutureExecutionTest
       end
 
       describe 'serializers' do
-        let(:save_and_load) do
```
adamruzicka: Are we sure that the `serialized_args` won't be changed by saving/loading it to/from a database?

iNecas: Ah, I first thouht this was a dead-code (missed the reference to it in the oirginal test), I've put it back

adamruzicka:
Looks good to me

iNecas:
@adamruzicka thanks for the review and coments!

PR #162, opened 2015-08-12
Support for canceling execution plans (inluding scheduled and plans with sub-plans)

iNecas:


iNecas:
@adamruzicka could you review it please?

lib/dynflow/scheduled_plan.rb
```diff
@@ -34,6 +34,12 @@ def error(message, history_entry = nil)
       execution_plan.root_plan_step.save
       execution_plan.execution_history.add history_entry, @world.id unless history_entry.nil?
       execution_plan.update_state :stopped
+      @world.persistence.delete_scheduled_plans(:execution_plan_uuid => execution_plan.id)
```
adamruzicka: I would suggest moving this line to the `cancel` method (or making additional changes to scheduler behavior).

When a scheduler finds an expired schedule record, it calls `error` on its execution plan and deletes the schedule record afterwards (along with other processed schedule records). This way expired schedule records would end up being deleted twice.

lib/dynflow/scheduled_plan.rb
```diff
@@ -34,6 +34,12 @@ def error(message, history_entry = nil)
...
+    end
+
+    def cancel
+      error("Scheudled task cancelled", "Scheduled task cancelled")
```
adamruzicka: Typo here

lib/dynflow/scheduled_plan.rb
iNecas: Updated

lib/dynflow/scheduled_plan.rb
iNecas: fixed

adamruzicka:
No more comments, ACK

iNecas:
Thanks for review @adamruzicka

PR #163, opened 2015-08-13
Remove active_support dependency

iNecas:
Don't force the users to use active support extensions, unless they want it on
their own. Dynflow now can operate without active support in place. On the other hand,
for backward compatibility, as well as user convenience (if the HashWithIndifferent
access is there, they probably are already used to it), if the active support is present,
use it's features instead of the alternative implementation.


adamruzicka:
Looks good, +1

iNecas:
Did tests against both environments with and without active record and everything seems to work as before.

PR #164, opened 2015-08-14
Keep pending subplans count in output

ares:
This is handy when you want to know how many tasks are pending and you don't want to calculate `total - success - failed` all the time


ares:
I'd be happy to test status just after trigger without subplans being executed if you tell me how. Not sure how to stop the execution so I can be sure that both will be pending at time of testing...

iNecas:
You can use this approach: https://github.com/Dynflow/dynflow/blob/master/test/action_test.rb#L449-L455, basically causing the subplans to "get stuck", checking the output, and then cancelling.

iNecas:
Thanks!

PR #171, opened 2015-10-08
Ensure the locks are released when world is being deleted

iNecas:
Otherwise, invalidated worlds still keep the locks after being deleted.

How to reproduce:

1. start first world: it should get the delayed executor flag. 2. invalidate
the world 3. start another world

Expected: the new world is flagged as delayed executor Before fix: the world
is not delayed executor, because an old DelayedExecutorLock is still acquired
for the old world.

### DIFF
--- a/lib/dynflow/coordinator.rb
+++ b/lib/dynflow/coordinator.rb
@@ -277,6 +277,7 @@ module Dynflow
 
     def delete_world(world)
       Type! world, Coordinator::ClientWorld, Coordinator::ExecutorWorld
+      release_by_owner("world:#{world.id}")
       delete_record(world)
     end
 
--- a/lib/dynflow/coordinator.rb
+++ b/lib/dynflow/coordinator.rb
@@ -155,6 +155,17 @@ module Dynflow
         @data[:world_id]
       end
 
+      def self.valid_owner_ids(coordinator)
+        coordinator.find_worlds.map { |w| "world:#{w.id}" }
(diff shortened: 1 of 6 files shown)


adamruzicka:
Works as expected, ACK

ares:
After applying the workaround from the [issue](http://projects.theforeman.org/issues/12088) I can confirm it works with the PR as expected.

lib/dynflow/coordinator.rb
```diff
@@ -285,5 +297,24 @@ def deactivate_world(world)
...
     end
+
+    def clean_orphaned_locks
+      cleanup_classes = [LockByWorld]
+      ret = []
+      cleanup_classes.each do |cleanup_class|
+        valid_owner_ids = cleanup_class.valid_owner_ids(self)
+        valid_classes = cleanup_class.valid_classes.map(&:name)
+        orphaned_locks = find_locks(class: valid_classes, exclude_owner_id: valid_owner_ids)
+        # reloadnig the valid owner ids to avoid race conditions
```
ares: s/reloadnig/reloading/

iNecas: updated

ares:
Reads good, not tested though.

adamruzicka:
Tested, works without applying the workaround.

iNecas:
Thanks @ares @adamruzicka for review and testing

PR #174, opened 2015-11-12
Extract transaction logic into a middleware

iNecas:
Together with ability to not use some middleware in some action, this
effective allows to turn off this functionality for some actions.


adamruzicka:
Code looks good, I'll test it tomorrow (or during the weekend...)

adamruzicka:
Did some testing, works nicely, ACK

iNecas:
@adamruzicka thanks for testing!

PR #177, opened 2015-12-04
Fixes #175 - update concurrent-ruby dependencies

### LINKED ISSUE #175: compatibility to newer concurrent-ruby version
There's the intention to update concurrent-ruby(-edge) in the FreeBSD ports tree: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=204874

Would it be possible to update dynflow to be compatible with 1.0?

iNecas:


pitr-ch:
:+1:

PR #178, opened 2015-12-07
We can add links to the navbar from outside of the dynflow

xprazak2:
It allows us to add links to the main navbar from outside of the dynflow (when dynflow console is mounted in a Rails application, for example). The parent application where console is mounted is beyond reach of console's routing, ~~so there is some url manipulation going on.~~
~~See [this](https://github.com/theforeman/foreman-tasks/pull/152) for usage.~~

### DIFF
--- a/lib/dynflow/web/navbar_items_helper.rb
+++ b/lib/dynflow/web/navbar_items_helper.rb
@@ -1,0 +1,16 @@
+module Dynflow
+  module Web
+    module NavbarItemsHelper
+      module Link
+
+        def self.link_back(url)
+          @@link_back ||= url
+        end
+      end
+
+      def back
+        Link.link_back request.referrer
+      end
+    end
+  end
+end
--- a/web/views/layout.erb
+++ b/web/views/layout.erb
@@ -36,6 +36,9 @@
             <ul class="nav navbar-nav">
               <li><a href="<%= url '/' %>">Execution plans</a></li>
               <li><a href="<%= url '/status' %>">Status</a></li>
+              <% if back %>
+                  <li><a href="<%= back %>">Back to Foreman</a></li>
+              <% end %>
             </ul>
             <ul class="nav navbar-nav navbar-right">
             </ul>
--- a/lib/dynflow/web.rb
+++ b/lib/dynflow/web.rb
@@ -18,6 +18,7 @@ module Dynflow
       File.join(web_dir, sub_dir)
     end
 
+    require 'dynflow/web/navbar_items_helper'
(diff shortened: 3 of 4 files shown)


xprazak2:
[test]

xprazak2:
The 'back' url gets noted only when accessing the console, then it is ignored. This way we can always go back to foreman even if we visit multiple pages in dynflow console.

xprazak2:
[test]

xprazak2:
[test]

adamruzicka:
If you restart foreman and access the dynflow console directly, then it links to `$host/foreman_tasks/dynflow` and stays this way if you access dynflow console later from tasks index until you restart foreman

lib/dynflow/web/navbar_items_helper.rb
```diff
@@ -0,0 +1,17 @@
+module Dynflow
+  module Web
+    module NavbarItemsHelper
+      module Link
+        @@link_back ||= nil
```
adamruzicka: `lhs ||= rhs` sets `lhs` to the value of `rhs` if `lhs` is `nil`. Setting it to `nil` if it is `nil` doesn't make much sense

xprazak2:
When accessing dynflow console directly after restart, the 'back' link is hidden. It will appear when we go through link from foreman_tasks list. This is a limitation when using `request.referrer`. If we are not ok with this and want the link to appear at all times, we probably need to abandon the dynamic generation of link based on the referrer.

web/views/layout.erb
```diff
@@ -36,6 +36,9 @@
             <ul class="nav navbar-nav">
               <li><a href="<%= url '/' %>">Execution plans</a></li>
               <li><a href="<%= url '/status' %>">Status</a></li>
+              <% if back %>
+                  <li><a href="<%= back %>">Back to Foreman</a></li>
```
iNecas: Given the simple logic, we could probably just use `request.referrer` here, right?

xprazak2: We could, but there are 2 problems:
1) User comes to dynflow console through the link in foreman_tasks list. Then clicks on anything in navbar and has no way of getting back to the foreman since `request.referrer` now points to task show page in dynflow.
2) Accessing the task in dynflow by url makes it point to itself. This could be possible source of confusion because 'Back to Foreman' now just refreshes the page.

ares:
Ping @inecas and @aruzicka. This seems useful and small enough, could we get this moving pls?

iNecas:
Closing in favor of #204

PR #179, opened 2016-01-18
Fixes #10751 - Concurrency level control

adamruzicka:
~~Global level is per-executor.~~

### DIFF
--- a/lib/dynflow/action/with_sub_plans.rb
+++ b/lib/dynflow/action/with_sub_plans.rb
@@ -17,8 +17,8 @@ module Dynflow
                end
              end),
             (on SubPlanFinished do
-              mark_as_done(event.execution_plan_id, event.success)
-              try_to_finish or suspend
+               mark_as_done(event.execution_plan_id, event.success)
+               try_to_finish or suspend
              end),
             (on Action::Cancellable::Cancel do
                cancel!
@@ -28,7 +28,15 @@ module Dynflow
     def initiate
       sub_plans = create_sub_plans
       sub_plans = Array[sub_plans] unless sub_plans.is_a? Array
-      wait_for_sub_plans(sub_plans)
+      if uses_concurrency_control
+        planned, failed = sub_plans.partition { |plan| plan.state == :planned }
+        calculate_time_distribution sub_plans.count
+        sub_plans = world.throttle_limiter.handle_plans! execution_plan_id,
+                                                         planned.map(&:id),
+                                                         failed.map(&:id),
+                                                         input[:concurrency_control]
+      end
+      wait_for_sub_plans sub_plans
     end
 
     # @abstract when the logic for the initiation of the subtasks
@@ -55,13 +63,40 @@ module Dynflow
     end
 
     def cancel!
+      @world.throttle_limiter.cancel!(execution_plan_id)
       sub_plans('state' => 'running').each(&:cancel)
       suspend
     end
 
     # Helper for creating sub plans
     def trigger(*args)
-      world.trigger { world.plan_with_caller(self, *args) }
+      if uses_concurrency_control
+        world.plan_with_caller(self, *args)
+      else
+        world.trigger { world.plan_with_caller(self, *args) }
+      end
+    end
+
+    def limit_concurrency_level(level)
+      input[:concurrency_control] ||= {}
+      input[:concurrency_control][:level] = ::Dynflow::Semaphores::Stateful.new(level).to_hash
+    end
+
+    def calculate_time_distribution(count)
+      time = input[:concurrency_control][:time]
+      unless time.nil? || time.is_a?(Hash)
+        # Assume concurrency level 1 unless stated otherwise
+        level = input[:concurrency_control].fetch(:level, {}).fetch(:free, 1)
+        semaphore = ::Dynflow::Semaphores::Stateful.new(nil, level,
+                                                        :interval => time.to_f / (count * level),
+                                                        :time_span => time)
+        input[:concurrency_control][:time] = semaphore.to_hash
+      end
+    end
+
+    def distribute_over_time(time_span)
+      input[:concurrency_control] ||= {}
+      input[:concurrency_control][:time] = time_span
     end
 
     def wait_for_sub_plans(sub_plans)
@@ -171,5 +206,8 @@ module Dynflow
       fail "A sub task failed" if output[:failed_count] > 0
     end
 
+    def uses_concurrency_control
+      @uses_concurrency_control = input.key? :concurrency_control
+    end
   end
 end
--- a/lib/dynflow/action/with_sub_plans.rb
+++ b/lib/dynflow/action/with_sub_plans.rb
@@ -31,10 +31,10 @@ module Dynflow
       if uses_concurrency_control
         planned, failed = sub_plans.partition { |plan| plan.state == :planned }
         calculate_time_distribution sub_plans.count
-        sub_plans = world.throttle_limiter.handle_plans! execution_plan_id,
+        sub_plans = world.throttle_limiter.handle_plans!(execution_plan_id,
                                                          planned.map(&:id),
                                                          failed.map(&:id),
-                                                         input[:concurrency_control]
+                                                         input[:concurrency_control])
       end
       wait_for_sub_plans sub_plans
     end
--- a/lib/dynflow/dispatcher/executor_dispatcher.rb
+++ b/lib/dynflow/dispatcher/executor_dispatcher.rb
@@ -1,7 +1,7 @@
 module Dynflow
   module Dispatcher
     class ExecutorDispatcher < Abstract
-      def initialize(world)
+      def initialize(world, semaphore)
         @world        = Type! world, World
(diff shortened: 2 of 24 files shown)


examples/middleware_concurrency_control.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+end
+
+
+class ConcurrencyControlExample < Dynflow::Action
+  include Dynflow::Action::WithSubPlans
+
+  middleware.use ::Dynflow::Middleware::Common::ConcurrencyControl
+
+  ConcurrencyLevel = 2
```
iNecas: `CAPITAL_LETTERS` for constants

examples/middleware_concurrency_control.rb
```diff
@@ -0,0 +1,75 @@
...
+  RunWithin = 2 * 60
+  
+  def plan(count)
+    limit_concurrency_level(ConcurrencyLevel)
+    distribute_over_time(RunWithin, count, ConcurrencyLevel)
+    super(:count => count)
+  end
+
+  def create_sub_plans
+    sleep 1
```
iNecas: why the sleep?

lib/dynflow/dispatcher/executor_dispatcher.rb
```diff
@@ -1,9 +1,10 @@
 module Dynflow
   module Dispatcher
     class ExecutorDispatcher < Abstract
-      def initialize(world)
+      def initialize(world, semaphore)
         @world        = Type! world, World
         @current_futures = Set.new
+        @semaphore = ::Dynflow::Semaphores::Aggregating.new(:global => semaphore)
```
iNecas: is this the global semafore? Shouldn't be part of the world (coordinator)?

lib/dynflow/middleware/common/concurrency_control.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+
+      private
+
+      def pass_available
+        pass ticket[:count => semaphore.drain]
+      end
+
+      def semaphore
+        semaphores ||= action.input[:concurrency_control].keys.reduce(Utils.indifferent_hash({})) do |acc, key|
```
iNecas: what `acc` stands for?

test/semaphores_test.rb
```diff
@@ -0,0 +1,99 @@
...
+    describe ::Dynflow::Semaphores::Aggregating do
+      let(:klass) { ::Dynflow::Semaphores::Aggregating }
+      let(:child_class) { ::Dynflow::Semaphores::Stateful }
+      let(:children) do
+        {
+          :child_A => child_class.new(3),
+          :child_B => child_class.new(2)
+        }
+      end
+      let(:assert_state) do
```
iNecas: maybe just method would be enough

lib/dynflow/action/with_sub_plans.rb
```diff
@@ -2,33 +2,58 @@ module Dynflow
...
+               start
+             end),
+            (on ExecutionTicket.(~any) do |count|
+               start count
              end),
             (on SubPlanFinished do
-              mark_as_done(event.execution_plan_id, event.success)
-              try_to_finish or suspend
+               mark_as_done(event.execution_plan_id, event.success)
+               try_to_finish or wake_n(event.execution_ticket.nil? ? 0 : event.execution_ticket.count) or suspend
```
iNecas: the method could be made shorter by extracting the bodies of contiions into separate methods. Especially this line would help being extracted into multiple lines

lib/dynflow/action/with_sub_plans.rb
```diff
@@ -2,33 +2,58 @@ module Dynflow
   module Action::WithSubPlans
```
iNecas: I wonder if the chnages on this action could be minimized, by defining some interface (methods the method needs to implement), that the middleware would call. Passing the execution tickets by modifying the events seems suboptimal. What if, instead of middleware, we would have some dependency injection here (just a method returning the semaphore-like object), that would help with the stuff? Let's talk about it on Monday

examples/middleware_concurrency_control.rb
adamruzicka: It is just a hacky workaround to have the plans' outputs printed after the description and webrick initialization notices

lib/dynflow/middleware/common/concurrency_control.rb
adamruzicka: accumulator

lib/dynflow/dispatcher/executor_dispatcher.rb
adamruzicka: The 'global' key might be a bit misleading, I had a bit different plans with this, we can talk about this more at the office.

lib/dynflow/action/with_sub_plans.rb
```diff
@@ -17,18 +17,19 @@ def run(event = nil)
...
                cancel!
              end)
     end
 
-    def initiate
+    def initiate(count = nil)
       sub_plans = create_sub_plans
       sub_plans = Array[sub_plans] unless sub_plans.is_a? Array
-      wait_for_sub_plans(sub_plans)
+      sub_plans = world.throttle_limiter.handle_plans!(execution_plan_id, sub_plans.map(&:id), input[:concurrency_control]) if uses_concurrency_control
```
iNecas: the line too long, if on a new line would make it more readable

lib/dynflow/semaphores/abstract.rb
```diff
@@ -0,0 +1,34 @@
+module Dynflow
+  module Semaphores
+    class Abstract
+      
+      def wait(thing)
```
iNecas: Some docs on what's expected form the methods to do would be nice.

lib/dynflow/throttle_limiter.rb
```diff
@@ -0,0 +1,105 @@
+module Dynflow
+  class ThrottleLimiter
```
iNecas: I wonder, if this could not be seen as different kind delayed executor?

examples/global_concurrency_control.rb
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env ruby
+
+example_description = <<DESC
+  Concurrency Control Example
+  ===================
+
+  This example shows, how to ConcurrencyControl middleware
```
iNecas: Make sure the remove the middleware mentions, as it seems we're not going the middleware way anymore.

iNecas:
I like the new appraoch much better! Looking foreward for more tests around it. But IMO the right move towards the final implementation!

lib/dynflow/semaphores/action_input.rb
```diff
@@ -0,0 +1,21 @@
+module Dynflow
+  module Semaphores
+    class ActionInput < Stateful
```
iNecas: Is the `ActionInput` the best name for the thing? It might be missleading and someone might thing it has something to do with `action.input`

lib/dynflow/throttle_limiter.rb
adamruzicka: Wouldn't it be odd to have something like `DelayedExecutor::ThrottleLimiter` which would not extend `DelayedExecutor::Abstract`? However I wouldn't mind moving it from the `::Dynflow` scope

iNecas: It would only make sense if the interface would match the
DelayedExecutor::Abstract (could be updated on both
DelayedExecution::Abstract and the TrhottleLimiter), but only if it was
obvious on how to do so. It can always be done later.

lib/dynflow/throttle_limiter.rb
```diff
@@ -0,0 +1,133 @@
...
+      def execute_triggered(triggered)
+        @world.execute(triggered.execution_plan_id, triggered.finished)
+      end
+
+      def set_up_clock_for(plan_id, initial = false)
+        if @semaphores[plan_id].children.key? :time
+          interval = @semaphores[plan_id].children[:time].meta[:interval]
+          timeout = @semaphores[plan_id].children[:time].meta[:time_span]
+          @world.clock.ping(self, interval, [:release, plan_id, :time])
+          @world.clock.ping(self, timeout, [:cancel, plan_id, 'timeout']) if initial
```
iNecas: Some more explanatory reason would be nice for the end user

test/concurrency_control_test.rb
```diff
@@ -0,0 +1,241 @@
...
+            check_step(plan, total, finished)
+          end
+          end_time = klok.current_time
+          (end_time - start_time).must_equal 4
+          world.throttle_limiter.observe(plan.id).must_equal []
+          world.throttle_limiter.core.ask!(:running).max.must_be :<=, level
+        end
+      end
+
+      # it 'fails tasks which failed to plan immediately' do
```
iNecas: commented code

iNecas:
If possible, prefer using method calls with ().

iNecas:
Tested and works nicely, Some minor comments.

iNecas:
Thanks @adamruzicka, works well. Nice job!

PR #180, opened 2016-01-28
Relax the concurrent ruby dependency

iNecas:
concurrent-ruby team has always been nice to us in terms of backward
compatibility: let's rely on sem-ver for now and relax the version requirements
a bit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language:
   - ruby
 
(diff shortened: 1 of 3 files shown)


adamruzicka:
ACK, tests run fine on my machine, apparently Travis still has [this issue](https://github.com/adamruzicka/dynflow/commit/ec987187122572294c60f26dd1d33171addc20c9#commitcomment-15557546)

PR #181, opened 2016-02-12
Fix world auto-termination

iNecas:
Due to too general rescue in concurrent ruby, we need to call the exit
in separate thread, otherwise the SystemExit error gets caught and
nothing happens.

Also, I've made the work with `@exit_on_terminate` thread-safe,
which should not be a case in MRI, but it's potentially unsynchronized
in future Ruby implementations.


iNecas:
The PR, that would fix the issue on concurrent ruby is here https://github.com/ruby-concurrency/concurrent-ruby/pull/507

adamruzicka:
Looks good, ACK

PR #183, opened 2016-04-26
The warnings in tests just made a mess form the the test output

iNecas:
And also made travis to fail on output size limit


ares:
LGTM

PR #185, opened 2016-05-11
Fix thread-safety issue with sequel iteration

iNecas:
The dataset.each is not thread-safe, as described in documentation for the
method

  https://github.com/jeremyevans/sequel/blob/4.20.0/lib/sequel/dataset/actions.rb#L126-L142

I suspect this to cause intermittent segmentation faults when under heaver
load.


adamruzicka:
Got to 4000th interation of your [reproducer](https://github.com/iNecas/dynflow/tree/segfault-reproducer) without encountering the issue, looks like the issue was solved. ACK

PR #188, opened 2016-06-15
Check plan is not lock before auto_execution

iNecas:
There is a potential race condition, where a plan is still in planned state,
but some executor has already locked it.

### DIFF
--- a/lib/dynflow/coordinator.rb
+++ b/lib/dynflow/coordinator.rb
@@ -146,6 +146,14 @@ module Dynflow
         @data.merge!(owner_id: "world:#{world.id}",  world_id: world.id)
       end
 
+      def self.lock_id(*args)
+        raise NoMethodError
+      end
+
+      def self.unique_filter(*args)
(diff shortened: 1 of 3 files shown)


lib/dynflow/coordinator.rb
```diff
@@ -146,6 +146,14 @@ def initialize(world)
         @data.merge!(owner_id: "world:#{world.id}",  world_id: world.id)
       end
 
+      def self.lock_id(*args)
+        riase NoMethodError
```
adamruzicka: typo

adamruzicka:
Any tips on how to trigger this race condition?

iNecas:
The easiest it to follow https://bugzilla.redhat.com/show_bug.cgi?id=1344707#c11

lib/dynflow/coordinator.rb
iNecas: Updated

adamruzicka:
Seems to really fix the issue, ACK

PR #190, opened 2016-07-08
Middleware support for present phase

iNecas:
This commit allows to hook in the action when the data are presented
to the user, for example to filter some sensitive information.


adamruzicka:
Works as expected, ACK

An example might be nice though...

iNecas:
I will write a middleware example in seperate PR.

PR #191, opened 2016-07-21
Adds Fail rescue strategy

adamruzicka:
- [x] tests

### DIFF
--- a/lib/dynflow/action/rescue.rb
+++ b/lib/dynflow/action/rescue.rb
@@ -2,7 +2,7 @@ module Dynflow
   module Action::Rescue
 
     Strategy = Algebrick.type do
-      variants Skip = atom, Pause = atom
+      variants Skip = atom, Pause = atom, Fail = atom
     end
 
     SuggestedStrategy = Algebrick.type do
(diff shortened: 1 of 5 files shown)


adamruzicka:
Test failures seem to be unrelated

lib/dynflow/action/rescue.rb
```diff
@@ -44,14 +44,20 @@ def rescue_strategy_for_planned_action(action)
...
-            suggested_strategies.all? { |suggested_strategy| suggested_strategy.strategy == Skip }
+      if suggested_strategies.empty?
         return Skip
       else
-        return Pause
+        # TODO: Find the safest rescue strategy among the suggested ones
+        if suggested_strategies.all? { |suggested_strategy| suggested_strategy.strategy == Skip }
+          return Skip
+        elsif suggested_strategies.any? { |suggested_strategy| suggested_strategy.strategy == Fail }
+          return Fail
```
iNecas: I'm not sure if one sub-action somewhere at the bottom of the plan tree should be able to set the whole task as failed. Notice that in Skip, all the sub-plans need to agree on the skip before proceeding, Also, the fail strategy is influencing also the steps that were not started yet. I think, for this strategy, should win only if all the pending steps are planned by the action that suggests the Fail strategy.

adamruzicka: Changed to the same behavior as used for `Skip`

adamruzicka:
We're green

iNecas:
Looks good, please squash so that we can merge it

adamruzicka:
@inecas squashed and tests are green

iNecas:
Thanks @adamruzicka!

PR #195, opened 2016-08-24
Add rubocop and fix whitespace issues

daviddavis:


test/action_test.rb
```diff
@@ -460,7 +460,7 @@ def run(event = nil)
           triggered_plan.finished.value.state.must_equal :stopped
           triggered_plan.finished.value.result.must_equal :success
         end
-     end
+      end
```
iNecas: Could you disable the cop here: the irregularity of the original one had just a bit higher artistic value than the new version.

iNecas: Just kidding :D

iNecas:
Thanks @daviddavis

PR #196, opened 2016-08-30
Fixes #16113 - Provide step data for presentation

adamruzicka:


lib/dynflow/execution_plan/steps/abstract.rb
```diff
@@ -112,8 +112,8 @@ def progress_weight
 
       # @return [Action] in presentation mode, intended for retrieving: progress information,
       # details, human outputs, etc.
-      def action(execution_plan)
-        world.persistence.load_action_for_presentation(execution_plan, action_id)
+      def action(execution_plan, with_step = false)
+        world.persistence.load_action_for_presentation(execution_plan, action_id, with_step ? self : nil)
```
iNecas: I wonder if we should not always include the step always there? Where it makes sence to not pass the step?

adamruzicka: Since it's the first case where the step data is needed, it seemed less intrusive to me to provide the step data on-demand instead of always, but I don't have strong opinions about this.

iNecas: If you can check the usages from other places to make sure we don't break somethink, and make it a new default, I think it's better to have one option: the `with_step` option doesn't make much sense to me right now.

adamruzicka:
It seems it is safe to always provide the step data, updated.

iNecas:
Thank you, tested and works well. merging

PR #197, opened 2016-09-02
Fix consistency check for stopped tasks

iNecas:
In occasional cases, it might happen that invalidated world got
marked as stopped before, while we try set it as paused, which
is not a valid transition.


adamruzicka:
ACK

PR #198, opened 2016-09-02
Don't use sequel map method

iNecas:
The Sequel's map method on dataset is not thread sefe, which might
occasionally lead to stack consistency issues. We already addreseed
similar case in 1510982b3d2b641ee4248cf6aa0ad9bd5392f828.

While it seemed it was the only place it might cause troubles,
similar thing could happen with auto-execute functionality,
when many tasks were pending. I've converted the rest of map
calls that return the result of map outside of the adapter to
make sure they are thread safe.


adamruzicka:
ACK

iNecas:
released in 0.8.15

PR #199, opened 2016-09-06
Use mediumtext instead of text for data columns in mysql

iNecas:


adamruzicka:
Do we have an issue for this or is this to avoid potentional issues in the future?

iNecas:
We got a report when trying  https://github.com/theforeman/foreman_ansible/pull/40 - the problem is mysql has different constraints for `:text` than other providers, therefore, we should use `:mediumtext` for all cases where we use `:text` otherwise, to keep the behavior the same

adamruzicka:
I wasn't aware of that report. ACK

dLobatog:
Works for me, no more exceptions on the data column anymore.

iNecas:
released in 0.8.15

PR #201, opened 2016-09-23
Fixes #16668 - don't crash when not successful for auto-execute

iNecas:


lib/dynflow/world.rb
```diff
@@ -360,6 +360,9 @@ def auto_execute
           end
         end.compact
       end
+    rescue Coordinator::LockError => e
+      logger.info "auto-executor lock already aquired: #{e.message}"
+      false
```
adamruzicka: I ran into an issue while running tests locally, where some tests (https://github.com/iNecas/dynflow/blob/57be15278e10e8c39da4300a30d0eb767ce4e3b3/test/abnormal_states_recovery_test.rb#L146) expect the `#auto_execute` to return an array of `wait`-able objects. It might happen we try to call `.each(&:wait)` on `false`. It needs to be handled either here or everywhere we use the return value

iNecas: Right, updated

adamruzicka:
ACK after Mr. Travis' permission

PR #203, opened 2016-09-26
Fixes #16690 - Pin pg gem for older rubies

adamruzicka:


adamruzicka:
Both #201 and #202 have failing tests because of this

PR #204, opened 2016-10-05
Fixes #12729 - Allow adding navbar links

adamruzicka:
Solves the same thing as https://github.com/Dynflow/dynflow/pull/178 differently


iNecas:
works as expected, merged, thanks @adamruzicka

PR #205, opened 2016-10-12
Fixes #16885 - use seuques' all instead of to_a

iNecas:
The `to_a` without `all` can still be thread unsafe at specific
circumstances.


adamruzicka:
APT

PR #206, opened 2016-11-02
Fixes #17177 - Handle invalidation with plans are missing

iNecas:
Under some circumstances (such as manually deleting execution plans in
database), it might happen that the execution plan we try to invalidate
is no longer present. This patch allows to recover from this kind of
situations.


adamruzicka:
Works as expected, APT

PR #207, opened 2016-11-02
When invalidating, pause only running steps

iNecas:
We hit an issue where we needed to invalidate a plan that was in planned
state. This lead to
 "invalid state transition planned >> paused in <Dynflow::ExecutionPlan"

It turns out the only valid transition to paused state is from running,
so switching the condition to just do it then.


adamruzicka:
Looks good to me, APT

PR #209, opened 2016-11-09
Refs #17283 - in thread executor for testing

iNecas:
Allows running the execution plan in side the same thread as the invoker.
Useful when relying on things like transactions in testing, where there is
no easy way to achieve that otherwise.

In order to reuse as much as possible in prepared
in-thread executor, as well as making better separation of
responsibilities, the parallel executor is now responsible only for the
execution itself, delegating the selection of the work items on newly
created director.


lib/dynflow/director/execution_plan_manager.rb
```diff
@@ -0,0 +1,109 @@
...
+        StepWorkItem.new(execution_plan.id, step).tap do |work|
+          @running_steps_manager.add(step, work)
+        end
+      end
+
+      # @return [Array<WorkItem>] of Work items to continue with
+      def what_is_next(work)
+        Type! work, WorkItem
+
+        compute_next_from_step =-> step do
```
adamruzicka: Is there a reason for having this as a lambda and not a regular method?

iNecas: I didn't do too much refactorints inside the methods, to not include too many changes at once, but I can update this one. Agree this is not very idiomatic ruby

lib/dynflow/director/execution_plan_manager.rb
```diff
@@ -0,0 +1,109 @@
...
+          work
+        when FinalizeWorkItem
+          raise unless @finalize_manager
+          finish
+        end
+      end
+
+      def event(event)
+        Type! event, Event
+        raise unless event.execution_plan_id == @execution_plan.id
```
adamruzicka: This (and three raises above) should probably have some kind of message, even if they function just as an assertion.

adamruzicka:
I inlined some comments and opened inecas/dynflow#3 for style issues I noticed in this PR.

iNecas:
@adamruzicka I've addressed the comments, ready for final review

adamruzicka:
Looks good to me, APT

PR #210, opened 2016-11-14
Refs #17283 - fix rescuing from InThreadExecutor

iNecas:
+ improvements in terminate interface of in thread world

### DIFF
--- a/lib/dynflow/executors/abstract.rb
+++ b/lib/dynflow/executors/abstract.rb
@@ -9,9 +9,13 @@ module Dynflow
         @logger = world.logger
       end
 
+      # @param execution_plan_id [Number] id of execution plan
+      # @param finished [Concurrent::Edge::Future]
+      # @param wait_for_acceptance [TrueClass|FalseClass] should the executor confirm receiving
+      # the event, disable if calling executor from within executor
       # @return [Concurrent::Edge::Future]
       # @raise when execution_plan_id is not accepted
-      def execute(execution_plan_id)
+      def execute(execution_plan_id, finished = Concurrent.future, wait_for_acceptance = true)
         raise NotImplementedError
       end
 
--- a/lib/dynflow/executors/abstract.rb
+++ b/lib/dynflow/executors/abstract.rb
@@ -9,7 +9,7 @@ module Dynflow
         @logger = world.logger
       end
 
-      # @param execution_plan_id [Number] id of execution plan
+      # @param execution_plan_id [String] id of execution plan
       # @param finished [Concurrent::Edge::Future]
       # @param wait_for_acceptance [TrueClass|FalseClass] should the executor confirm receiving
       # the event, disable if calling executor from within executor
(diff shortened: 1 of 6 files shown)


lib/dynflow/executors/abstract.rb
```diff
@@ -9,9 +9,13 @@ def initialize(world)
         @logger = world.logger
       end
 
+      # @param execution_plan_id [Number] id of execution plan
```
adamruzicka: Isn't `execution_plan_id` a string?

test/testing_test.rb
```diff
@@ -213,11 +213,32 @@ def plan_input
         end
       end
 
-      it "is able to handle when events ===" do
+      it "is able to handle when events" do
```
adamruzicka: How is this test different from [this one](https://github.com/iNecas/dynflow/blob/21173d948e3fc40fc5d05015595240fc738de656/test/testing_test.rb#L204)?

test/testing_test.rb
```diff
@@ -213,11 +213,32 @@ def plan_input
...
+
+      describe 'auto rescue' do
+        let(:world) do
+          WorldFactory.create_world(Dynflow::Testing::InThreadWorld) do |config|
+            config.auto_rescue = true
+          end
+        end
+
+        describe 'of plan with skips' do
+
```
adamruzicka: This empty line probably shouldn't be here

lib/dynflow/executors/abstract.rb
iNecas: It is! :)

test/testing_test.rb
iNecas: Apparently, Ruby is not as smart as it looks like, and can't tell what  I want to do just from the string description. I forgot to update the plan I wanted to actually test, and it turned out I had some issues with the event handling in the in thread executor, that should now be solved

test/testing_test.rb
iNecas: done

iNecas:
Thanks @adamruzicka for very valuable review! The things should be fixed now

lib/dynflow/testing/in_thread_executor.rb
```diff
@@ -7,16 +7,16 @@ def initialize(world)
         @work_items = Queue.new
       end
 
-      def execute(execution_plan_id, finished)
+      def execute(execution_plan_id, finished, _wait_for_acceptance = true)
```
adamruzicka: Shouldn't there be a default value for `finished` to have the interface exactly the same as the `abstract` and `parallel` executors have?

iNecas: Updated

adamruzicka:
Alright, let's see what Mr. Travis has to say and get this in

ares:
Mr. Travis is unhappy on just one worker, hopefully unrelated.

adamruzicka:
We're green now

PR #211, opened 2016-11-22
Fixes #17175 - Added watcher that can terminate a world

ShimShtein:
This watcher polls the memory size of the process it runs in and terminates the world if memory limit is exceeded.
This commit also adds `terminated` event to `World` class, so anyone can add a listener that will perform tasks after a world is terminated.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ end
 
 group :pry do
   gem 'pry'
+  gem 'pry-byebug'
 end
 
 group :postgresql do
@@ -28,4 +29,8 @@ end
 
 group :lint do
   gem 'rubocop', '0.39.0'
+end
+
+group :memory_watcher do
+  gem 'get_process_mem'
 end
--- a/lib/dynflow/watchers/memory_consumption_watcher.rb
+++ b/lib/dynflow/watchers/memory_consumption_watcher.rb
@@ -1,0 +1,32 @@
+require 'get_process_mem'
+
+module Dynflow
+  module Watchers
+    class MemoryConsumptionWatcher
+
+      attr_reader :memory_limit, :world
+
+      def initialize(world, memory_limit, options)
+        @memory_limit = memory_limit
+        @world = world
+        @polling_interval = options[:polling_interval] || 60
+        @memory_info_provider = options[:memory_info_provider] || GetProcessMem.new
+        set_timer options[:initial_wait] || @polling_interval
+      end
+
+      def check_memory_state
+        if @memory_info_provider.bytes > @memory_limit
+          # terminate the world and stop polling
+          world.terminate
+        else
+          # memory is under the limit - keep waiting
+          set_timer
+        end
+      end
+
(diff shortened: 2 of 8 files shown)


lib/dynflow/watchers/memory_consumption_watcher.rb
```diff
@@ -0,0 +1,32 @@
...
+  module Watchers
+    class MemoryConsumptionWatcher
+
+      attr_reader :memory_limit, :world
+
+      def initialize(world, memory_limit, options)
+        @memory_limit = memory_limit
+        @world = world
+        @polling_interval = options[:polling_interval] || 10
+        @memory_watcher = options[:memory_watcher] || GetProcessMem.new
```
iNecas: `s/memory_watcher/memory_info_loader` or somethng similar, the watcher is this class

lib/dynflow/watchers/memory_consumption_watcher.rb
```diff
@@ -0,0 +1,32 @@
+require 'get_process_mem'
+
+module Dynflow
+  module Watchers
+    class MemoryConsumptionWatcher
+
+      attr_reader :memory_limit, :world
+
+      def initialize(world, memory_limit, options)
```
iNecas: I would suggest to have an option to have a minimal limit for the world to live (we would need to set @started_at variable in the world, to be able to check this)

The reason would be to prevent immediate termination after the executor starts, without giving it a change to actually do something - this would happen if the limit would be set too low.

iNecas:
We are watching if we mark all futures/events as finished at the end, this was missing https://github.com/ShimShtein/dynflow/pull/1

iNecas:
Nitpick: there should be an empty line between subject and body in commit message :)

Gemfile
```diff
@@ -29,3 +30,7 @@ end
 group :lint do
   gem 'rubocop', '0.39.0'
 end
+
+group :memory_watcher do
+  gem 'get_process_mem'
```
ShimShtein: Didn't want to add it to gemspec, since it's optional.

lib/dynflow/watchers/memory_consumption_watcher.rb
ShimShtein: This is why I have added `options[:initial_wait]` you can pass a longer value in order to give the executor the time it needs to start.

ShimShtein:
Fixed everything except moving to gemspec. Now the tests can run.

Gemfile
iNecas: I see

iNecas:
Technically I approve this, but let's wait for the foreman-tasks portion of the work to test it together, in case we hit some issue here

ShimShtein:
Added an example and fixed a small bug that was not covered by the tests.

iNecas:
Thanks @ShimShtein , I still need to do final testing in foreman-tasks, but this part should be good to go.

iNecas:

PR #214, opened 2017-02-09
Batch planning of sub tasks

adamruzicka:
### Quickstart Guide
- include the `::Dynflow::Action::WithBulkSubPlans` module into the action.
- implement `total_count` method returning the total count of plans we want to create
- implement `entries(FROM, SIZE)` method returning the batch starting at position `FROM` and spanning `SIZE` items
- the `BATCH_SIZE` constant can be redefined to change the size of the batch
- use `current_batch` inside `create_sub_plans` to get the current batch

### TODO
- [X] resuming paused tasks
- [X] tests
- [x] fix cancelling
- [ ] fix concurrency control after restart

### DIFF
--- a/lib/dynflow/action/with_bulk_sub_plans.rb
+++ b/lib/dynflow/action/with_bulk_sub_plans.rb
@@ -1,0 +1,54 @@
+module Dynflow
+  module Action::WithBulkSubPlans
+    include Dynflow::Action::Cancellable
+
+    BATCH_SIZE=10
+
+    # Should return a slice of size items starting from item with index from
+    def entries(from, size)
+      raise NotImplementedError
+    end
+
+    def total_count
+      raise NotImplementedError
+    end
+
+    def current_batch
+      start_position = output[:total_count]
+      size = start_position + BATCH_SIZE > total_count ? total_count - start_position : BATCH_SIZE
+      entries(start_position, size)
+    end
+
+    def done?
+      super && total_count == output[:total_count]
+    end
+
+    def resume
+      # TODO
+      super
+    end
+
+    def run_progress
+      if counts_set? && output[:total_count] > 0
+        (output[:success_count] + output[:failed_count]).to_f / total_count
+      else
+        0.1
+      end
+    end
+
+    def try_to_finish
+      if can_spawn_next_batch?
+        spawn_plans
+      else
+        super
+      end
+    end
+
+    private
+
+    def can_spawn_next_batch?
+      total_count > output[:total_count]
+    end
+
+  end
+end
--- a/lib/dynflow/action/with_bulk_sub_plans.rb
+++ b/lib/dynflow/action/with_bulk_sub_plans.rb
@@ -9,6 +9,17 @@ module Dynflow
       raise NotImplementedError
     end
 
+    PlanNextBatch = Algebrick.atom
+
+    def run(event = nil)
+      if event === PlanNextBatch
+        spawn_plans if can_spawn_next_batch?
+        suspend
+      else
+        super
+      end
+    end
+
     # Should return the expected total count of tasks
     def total_count
       raise NotImplementedError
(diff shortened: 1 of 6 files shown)


lib/dynflow/action/with_bulk_sub_plans.rb
```diff
@@ -0,0 +1,54 @@
...
+      super && total_count == output[:total_count]
+    end
+
+    def resume
+      # TODO
+      super
+    end
+
+    def run_progress
+      if counts_set? && output[:total_count] > 0
```
iNecas: Shouldn't we use `total_count` method in the condition?

lib/dynflow/action/with_bulk_sub_plans.rb
```diff
@@ -0,0 +1,54 @@
+module Dynflow
+  module Action::WithBulkSubPlans
+    include Dynflow::Action::Cancellable
+
+    BATCH_SIZE=10
+
+    # Should return a slice of size items starting from item with index from
+    def entries(from, size)
```
iNecas: What about calling this method simply `batch`?

iNecas:
ACK on the approach. Please proceed with finishing the PRs

lib/dynflow/action/with_bulk_sub_plans.rb
adamruzicka: sounds good

lib/dynflow/action/with_bulk_sub_plans.rb
adamruzicka: The condition was there just to refrain from dividing by zero, this shouldn't happen anymore, removed it

iNecas:
The builds are failing on rubocop

iNecas:
While testing, I've noticed the current implementation plans the next batch once some plans finished: we can do better just by making sure we send the event to plan next batch sooner https://gist.github.com/iNecas/4ad20c1451f46ef91b198ed5e9fdd413

adamruzicka:
Cancelling should now work properly.

I'm afraid the concurrency control after executor's restart never worked. I'll create a new PR for it

lib/dynflow/action/with_bulk_sub_plans.rb
```diff
@@ -0,0 +1,83 @@
...
+    end
+
+    # Should return the expected total count of tasks
+    def total_count
+      raise NotImplementedError
+    end
+
+    # Returns the items in the current batch
+    def current_batch
+      start_position = output[:total_count]
```
iNecas: I suggest keeping the total_count in output to match the `total_count` method, and introduce `planned_count` to keep the number of tasks we have planned: The `total_count` is misleading.

lib/dynflow/action/with_bulk_sub_plans.rb
```diff
@@ -0,0 +1,83 @@
+module Dynflow
+  module Action::WithBulkSubPlans
+    include Dynflow::Action::Cancellable
+
+    BATCH_SIZE = 10
```
iNecas: Let's rename this to `DEAFAULT_BATCH_SIZE`, as it can be changed by overriding `batch_size` method. Also, I would recommand making default a bit higher (even 100) should be safe for most cases, when not doing too much in the planning phase.

lib/dynflow/action/with_sub_plans.rb
```diff
@@ -26,16 +26,20 @@ def run(event = nil)
     end
 
     def initiate
-      sub_plans = create_sub_plans
-      sub_plans = Array[sub_plans] unless sub_plans.is_a? Array
+      output.update(:total_count   => 0,
```
iNecas: Do we need to reset the numbers here? It would break the `done?` method when called sooner than any planning happens. When we go with `planned_count` number, we should not need to reset this numbers at the beginning (just initiating the `planned_count` in the `WithBulkSubPlans`

adamruzicka:
Updated

iNecas:
Tests are faillnig

iNecas:
There are still some tasks/rex changes to be made, but this one should be good to go. Thanks @adamruzicka

iNecas:

PR #216, opened 2017-02-22
ActiveJob adapter

dLobatog:
This commit adds an adapter to allow Dynflow to be used as an ActiveJob
backend. After this, jobs in Rails can be defined as ActiveJob::Base,
enqueued with "SampleJob.perform_later" or scheduled with
"SampleJob.set(:wait => 10.seconds).perform_later('abc')"

https://gist.github.com/anonymous/429e7171e867c94182839a4b0d1ddddc

### DIFF
--- a/dynflow.gemspec
+++ b/dynflow.gemspec
@@ -29,7 +29,8 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rack-test"
   s.add_development_dependency "minitest"
   s.add_development_dependency "minitest-reporters"
-  s.add_development_dependency "activerecord"
+  s.add_development_dependency "activerecord", '< 5.0.0'
+  s.add_development_dependency 'activejob', '< 5.0.0'
   s.add_development_dependency "sqlite3"
   s.add_development_dependency "sinatra"
 end
--- a/lib/dynflow.rb
+++ b/lib/dynflow.rb
@@ -49,4 +49,16 @@ module Dynflow
   require 'dynflow/semaphores'
   require 'dynflow/throttle_limiter'
   require 'dynflow/config'
+
+  if defined? ::ActiveJob
+    require 'dynflow/active_job/queue_adapter'
+
+    class Railtie < Rails::Railtie
+      config.before_initialize do
+        ::ActiveJob::QueueAdapters.include(
+          Dynflow::ActiveJob::QueueAdapters
+        )
+      end
+    end
+  end
 end
--- a/lib/dynflow/active_job/queue_adapter.rb
+++ b/lib/dynflow/active_job/queue_adapter.rb
@@ -1,0 +1,34 @@
+module Dynflow
+  module ActiveJob
+    module QueueAdapters
+      # To use Dynflow, set the queue_adapter config to +:dynflow+.
+      #
+      #   Rails.application.config.active_job.queue_adapter = :dynflow
+      class DynflowAdapter
+        extend ActiveSupport::Concern
+
+        class << self
+          def enqueue(job)
+            Rails.application.dynflow.world.trigger(JobWrapper, job.serialize)
+          end
+
+          def enqueue_at(job, timestamp)
+            Rails.application.dynflow.world.delay(JobWrapper, { :start_at => Time.at(timestamp) }, job.serialize)
+          end
+        end
+      end
+
+      class JobWrapper < Dynflow::Action
+        def plan(attributes)
+          input[:job_class] = attributes['job_class']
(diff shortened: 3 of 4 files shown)


dynflow.gemspec
```diff
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rack-test"
   s.add_development_dependency "minitest"
   s.add_development_dependency "minitest-reporters"
-  s.add_development_dependency "activerecord"
+  s.add_development_dependency "activerecord", '< 5.0.0'
```
ohadlevy: out of curiosity, why do we need development dependencies?

iNecas:
Looks good, there are some issues with pinned concurrent-ruby gems, I will look into it today, could you add some tests in the meantime for this code? Having active job a developmnet dependency for running tests is ok

lib/dynflow/active_job/queue_adapter.rb
```diff
@@ -0,0 +1,34 @@
+module Dynflow
+  module ActiveJob
+    module QueueAdapters
+      # To use Dynflow, set the queue_adapter config to +:dynflow+.
```
ohadlevy: would this file eventually send as a pr to active job project?

lib/dynflow/active_job/queue_adapter.rb
```diff
@@ -0,0 +1,34 @@
...
+      #
+      #   Rails.application.config.active_job.queue_adapter = :dynflow
+      class DynflowAdapter
+        extend ActiveSupport::Concern
+
+        class << self
+          def enqueue(job)
+            Foreman.dynflow.world.trigger(JobWrapper, job.serialize)
```
ohadlevy: should we really say Foreman here? is dynflow not usable outside the foreman scope?

iNecas: Good point: the world to be used needs to be configured somewhere

iNecas:
For test failures: it seems bundler has some issues resolving dependencies lately (it worked fine with older bundler versions), needed to pin concurrent-ruby-ext to specific version in ihttps://github.com/Dynflow/dynflow/pull/217/files, I will work towards updateding to later versions of concurrent-ruby-edge, but that will be a bit larger change that I don't want to block others on

lib/dynflow/active_job/queue_adapter.rb
dLobatog: ActiveJob is part of Rails - I would expect so, yes

lib/dynflow/active_job/queue_adapter.rb
```diff
@@ -0,0 +1,34 @@
...
+    module QueueAdapters
+      # To use Dynflow, set the queue_adapter config to +:dynflow+.
+      #
+      #   Rails.application.config.active_job.queue_adapter = :dynflow
+      class DynflowAdapter
+        extend ActiveSupport::Concern
+
+        class << self
+          def enqueue(job)
+            Rails.application.dynflow.world.trigger(JobWrapper, job.serialize)
```
dLobatog: Substituted `Foreman` by `Rails.application` - sorry!

dLobatog:
@iNecas This has tests now

dynflow.gemspec
dLobatog: For testing the parts that integrate with these optional gems

ohadlevy: there is no s.add_testing_dependency? :)

dLobatog: http://guides.rubygems.org/specification-reference/ that doesn't seem to exist, unfortunately

lib/dynflow.rb
```diff
@@ -49,4 +49,16 @@ class Error < StandardError
   require 'dynflow/semaphores'
   require 'dynflow/throttle_limiter'
   require 'dynflow/config'
+
+  if defined? ::ActiveJob
+    require 'dynflow/active_job/queue_adapter'
+
+    class Railtie < Rails::Railtie
```
ohadlevy: must AJ means Rails is defined too?

dLobatog: I could not find any example of ActiveJob usage without Rails - so I don't think they support it to work without it, it's even in the same repository as Rails (rails/rails). The reason why I check for `::ActiveJob` is in case someone is using Dynflow with some Rails version that does not have ActiveJob.

iNecas:
Ok, I don't see anything that should prevent us form merging: unless anyone objects, I will do it in few hours

iNecas:
Thanks @dLobatog!

PR #217, opened 2017-02-22
Pin concurrent-ruby-ext to 1.0.3

iNecas:
New bundler has issues resolving more complicated dependencies,
pinning the ext we use for testing to 1.0.3 for now.


adamruzicka:
APT

PR #218, opened 2017-02-23
Rails initialization

dLobatog:
I made this PR depend on https://github.com/Dynflow/dynflow/pull/216 - so let's try to get them in order to reduce the diff a little. You should be familiar with most of this code. Check out its usage on https://github.com/theforeman/foreman/pull/4316#discussion_r102676503

### DIFF
--- a/lib/dynflow.rb
+++ b/lib/dynflow.rb
@@ -61,4 +61,8 @@ module Dynflow
       end
     end
   end
+
+  if defined? Rails
+    require 'dynflow/rails'
+  end
 end
--- a/lib/dynflow/rails/configuration.rb
+++ b/lib/dynflow/rails/configuration.rb
@@ -1,0 +1,146 @@
+module Dynflow
+  class Rails
+    class Configuration
+      # the number of threads in the pool handling the execution
+      attr_accessor :pool_size
+
+      # the size of db connection pool
+      attr_accessor :db_pool_size
+
+      # set true if the executor runs externally (by default true in procution, othewise false)
+      attr_accessor :remote
+      alias remote? remote
+
+      # what transaction adapater should be used, by default, it uses the ActiveRecord
+      # based adapter, expecting ActiveRecord is used as ORM in the application
+      attr_accessor :transaction_adapter
+
+      attr_accessor :eager_load_paths
+
+      attr_accessor :lazy_initialization
+
+      # what rake tasks should run their own executor, not depending on the external one
+      attr_accessor :rake_tasks_with_executor
+
+      def initialize
+        self.pool_size                = 5
+        self.db_pool_size             = pool_size + 5
+        self.remote                   = ::Rails.env.production?
+        self.transaction_adapter      = ::Dynflow::TransactionAdapters::ActiveRecord.new
+        self.eager_load_paths         = []
+        self.lazy_initialization      = !::Rails.env.production?
+        self.rake_tasks_with_executor = %w(db:migrate db:seed)
+
+        @on_init = []
+      end
+
+      # Action related info such as exceptions raised inside the actions' methods
+      # To be overridden in the Rails application
+      def action_logger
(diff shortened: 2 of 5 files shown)


lib/dynflow.rb
```diff
@@ -49,4 +49,20 @@ class Error < StandardError
   require 'dynflow/semaphores'
   require 'dynflow/throttle_limiter'
   require 'dynflow/config'
+
+  if defined? ::ActiveJob
+    require 'dynflow/active_job/queue_adapter'
+
+    class Railtie < Rails::Railtie
```
ohadlevy: should this be moved into the rails section?

lib/dynflow/rails/configuration.rb
```diff
@@ -0,0 +1,151 @@
...
+
+      attr_accessor :lazy_initialization
+
+      # what rake tasks should run their own executor, not depending on the external one
+      attr_accessor :rake_tasks_with_executor
+
+      # If true, the ForemanTasks::Concerns::ActionTriggering will make
+      # no effect. Useful for testing, where we mignt not want to execute
+      # the orchestration tied to the models.
+      attr_accessor :disable_active_record_actions
```
iNecas: This is not valid for pure dynfow

lib/dynflow/rails/persistence.rb
```diff
@@ -0,0 +1,26 @@
+module Dynflow
```
iNecas: I don't think we need this file without foreman-tasks

lib/dynflow/rails/daemon.rb
```diff
@@ -0,0 +1,72 @@
...
+        sleep
+      ensure
+        STDOUT.puts('Exiting')
+      end
+
+      # run the executor as a daemon
+      def run_background(command = 'start', options = {})
+        default_options = { rails_root: Dir.pwd,
+                            process_name: 'dynflow_executor',
+                            pid_dir: File.join(::Rails.root, '/tmp/pids'),
```
ohadlevy: I think it should be
```ruby
  pid_dir: File.join(::Rails.root, 'tmp','pids'),
```

lib/dynflow/rails/daemon.rb
```diff
@@ -0,0 +1,72 @@
...
+                            log_dir: File.join(::Rails.root, 'log'),
+                            wait_attempts: 300,
+                            wait_sleep: 1,
+                            executors_count: (ENV['EXECUTORS_COUNT'] || 1).to_i }
+        options = default_options.merge(options)
+        FileUtils.mkdir_p(options[:pid_dir])
```
ohadlevy: not sure if its required, but i guess not harmful?

lib/dynflow/rails/configuration.rb
dLobatog: OK - I suppose the tasks Configuration object could introduce it

lib/dynflow/rails/persistence.rb
dLobatog: Indeed, I'll rewrite it in Foreman-tasks so it's Configuration file overrides initialize_persistence with one that actually saves stuff like this.

dLobatog:
@iNecas Updated to remove the tasks option and the PersistenceAdapter for tasks. -tasks will simply override Configuration to define its own persistence model based on Sequel. Pretty similar to how Foreman does it: (https://github.com/theforeman/foreman/pull/4316/files#diff-b227cab047536b0261bee94c657af2deR2).

dLobatog:
Updated to fix 1 small conflict with the ActiveJob provider 😄

iNecas:
Thanks @dLobatog, I'm releasing new version now so that you can proceed with other parts of your effort

iNecas:
dynflow-0.8.20 is all yours @dLobatog :)

PR #220, opened 2017-03-09
Add label to execution_plan

iNecas:
The value of the label is determined by the `action#label`. The default
goes to the `class.name` but can be overriden, for example ActiveJob can
override it with the job class name.

### DIFF
--- a/lib/dynflow/active_job/queue_adapter.rb
+++ b/lib/dynflow/active_job/queue_adapter.rb
@@ -28,6 +28,10 @@ module Dynflow
         def run
           input[:job_class].constantize.perform_now(*input[:job_arguments])
         end
+
+        def label
+          input[:job_class]
+        end
       end
     end
   end
--- a/lib/dynflow/action.rb
+++ b/lib/dynflow/action.rb
@@ -1,4 +1,5 @@
 module Dynflow
+  # rubocop:disable Metrics/ClassLength
   class Action < Serializable
 
     OutputReference = ExecutionPlan::OutputReference
@@ -120,6 +121,10 @@ module Dynflow
     def phase!(*phases)
       phase?(*phases) or
         raise TypeError, "Wrong phase #{phase}, required #{phases}"
+    end
+
+    def label
+      self.class.name
     end
 
     def input=(hash)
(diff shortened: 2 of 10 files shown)


lib/dynflow/active_job/queue_adapter.rb
```diff
@@ -28,6 +28,10 @@ def plan(attributes)
         def run
           input[:job_class].constantize.perform_now(*input[:job_arguments])
         end
+
+        def label
+          input[:job_class]
```
dLobatog: I think it'd be nice to have `job_class` or `input[:job_class].constantize.label`. The reason is that if one has 2 jobs scheduled, for example SyncLDAPUsers(usergroup), I want to check on initialization if there's a job scheduled for that already. 

```
Dynflow.world.find_execution_plans :label => "SyncLDAPUser" 
```

is not good enough, as it assumes only 1 job will be scheduled of the same class at any point in time.

```
Dynflow.world.find_execution_plans :label => "sync-ldap-user-myfavoritegroup" 
```

is different and does not make that assumption (it allows me to iterate over all usergroups and only schedule syncs for the ones that are not scheduled). 

It'd be as simple as defining the label in the ActiveJob as 

```
def label
    "sync-ldap-users-#{@usergroup}"
end
```

iNecas: AFAIK you pass the args to the `perform` method of the active job: I don't see where in the active job class you would be able to define the label.

The label here is just do disctinguish differend kinds of jobs, but for your case, I believe there should be something on application layer to link the resources and jobs. We use foreman-tasks for it right now though locks. I believe we need a counterpart in foreman.

iNecas:
@dLobatog I believe there is not more I can do here to support your case and that the current patch is good enough to provide you ability to achieve the recurring check. 

@adamruzicka would you mind having a look?

adamruzicka:
Looks good and works as expected.

I noticed you added some `#rubocop:disable` and never enabled them again. Was it intentional?

iNecas:
Yes, it was intentional.

iNecas:
I'm merging this, as it will help us also performance wise. Let's leave other comments for further PRs

iNecas:

PR #222, opened 2017-03-21
Core dump tools look for core terminated.

bkearney:
Added some more text to make it look less like an error.

### DIFF
--- a/lib/dynflow/executors/parallel/core.rb
+++ b/lib/dynflow/executors/parallel/core.rb
@@ -43,13 +43,13 @@ module Dynflow
 
         def start_termination(*args)
           super
-          logger.info 'shutting down Core ...'
+          logger.info 'shutting down Dynflow Core ...'
           @pool.tell([:start_termination, Concurrent.future])
         end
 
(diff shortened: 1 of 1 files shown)


lib/dynflow/executors/parallel/core.rb
```diff
@@ -43,13 +43,13 @@ def handle_persistence_error(error)
...
           super
-          logger.info 'shutting down Core ...'
+          logger.info 'shutting down Dynflow Core ...'
           @pool.tell([:start_termination, Concurrent.future])
         end
 
         def finish_termination
           @director.terminate
-          logger.error '... core terminated.'
+          logger.error '... Dynflow Core has terminated.'
```
iNecas: Good point, maybe even 'terminate' is not the right word. What about 'Dynflow Core has shut down' or something more positive :)

bkearney: @iNecas I repushed with the use of shutdown.

iNecas:
Thanks @bkearney

PR #224, opened 2017-03-27
Add docs for how to use subplans

akofink:
I'm very open to any feedback on this one.

### DIFF
--- a/doc/pages/source/documentation/index.md
+++ b/doc/pages/source/documentation/index.md
@@ -950,7 +950,27 @@ to the chain of middleware execution.
 ### Sub-plans
 
 -   *when to use?*
--   *how to use?*
+
+To use sub-plans, you must include the `Dynflow::Action::WithSubPlans` module
+and override the `create_sub_plans` method. Inside the `create_sub_plans`
+method, you use the `trigger` method to create sub-tasks that will be executed
+in no particular order during the run phase. The parent task will wait for the
+sub-tasks to finish without blocking a thread in a pool while waiting.
+
+```rb
+class MyAction < Actions::EntryAction
+  include Dynflow::Action::WithSubPlans
+
+  ...
+
+  def create_sub_plans
+    [
+      trigger(Actions::OtherAction, action_param1, action_opts),
+      trigger(Actions::AnotherAction)
+    ]
+  end
+end
+```
 
 ## How it works TODO
 


doc/pages/source/documentation/index.md
```diff
@@ -950,7 +950,24 @@ to the chain of middleware execution.
 ### Sub-plans
 
 -   *when to use?*
--   *how to use?*
+
+To use sub-plans, you must include the `Dynflow::Action::WithSubPlans` module
+and override the `create_sub_plans` method. Inside the `create_sub_plans`
+method, you use the `trigger` method to create sub-tasks that will be executed
+in sequence but will not block the parent task.
```
adamruzicka: > you use the `trigger` method to create sub-tasks that will be executed in sequence

That's not entirely correct. The spawned sub-tasks are usually planned in sequence, but there are no other constraints on how they are actually executed (the run and finalize phase) and because of that we can't really say they're "executed in sequence".

doc/pages/source/documentation/index.md
```diff
@@ -950,7 +950,24 @@ to the chain of middleware execution.
...
+
+```rb
+class MyAction < Actions::EntryAction
+  include Dynflow::Action::WithSubPlans
+
+  ...
+
+  def create_sub_plans
+    trigger(Actions::OtherAction, action_param1, action_opts)
+    trigger(Actions::AnotherAction)
```
adamruzicka: The parent task expects this method to return either a single task (in case it has only one sub-task) or an array of them (in case of multiple sub-tasks). In this example the parent task would lose track of `OtherAction` because it is not returned from the method.

adamruzicka:
@akofink Thank you for opening this, docs are always welcomed. If you need any information or some questions answered feel free to ping me anytime. Left some comments inline

doc/pages/source/documentation/index.md
akofink: So, if you trigger two sub-tasks, in the same `create_sub_plans` method, the second task isn't guaranteed to start after the first task is finished?

adamruzicka: Basically yes

doc/pages/source/documentation/index.md
```diff
@@ -950,7 +950,27 @@ to the chain of middleware execution.
 ### Sub-plans
 
 -   *when to use?*
--   *how to use?*
+
+To use sub-plans, you must include the `Dynflow::Action::WithSubPlans` module
+and override the `create_sub_plans` method. Inside the `create_sub_plans`
+method, you use the `trigger` method to create sub-tasks that will be executed
+in no particular order during the run phase. These sub-tasks will not block the
+parent task.
```
iNecas: This is not precise: the parent task will still wait for the sub-plans to finish. The important thing is it's not blocking a thread in a pool while waiting. Mentioning that the reason for this is busy-waiting for the task in run method could lead to dead-locks, when all threads would be consumed by such a waiting, no tasks could be finished.

akofink:
Updated

iNecas:
Thanks @akofink

PR #227, opened 2017-04-03
Fixes #19146 - handle invalid plans when invalidating worlds

iNecas:
In a corner case, the data can be inconsistent in a way, that prevents
the world to be invalidate properly. Although this is usually a result
of unsafe manipulation with data on database level, we need to try
to resolve the situation to at least keep clean the runtime data properly,
as there is not much more we can do about it.

### DIFF
--- a/lib/dynflow/errors.rb
+++ b/lib/dynflow/errors.rb
@@ -31,6 +31,9 @@ module Dynflow
       end
     end
 
+    class DataConsistencyError < Dynflow::Error
+    end
+
     class PersistenceError < Dynflow::Error
       def self.delegate(original_exception)
         self.new("caused by #{original_exception.class}: #{original_exception.message}").tap do |e|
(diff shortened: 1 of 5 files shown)


beav:
I'm not in dynflow project, but change looks good to me.

iNecas:
@adamruzicka mind to have a look?

adamruzicka:
@iNecas The immediate issue is fixed (the world gets invalidated properly), however the "broken" execution plans are still left behind and accessing dynflow index raises `Dynflow::Errors::DataConsistencyError at /
Could not load steps for execution plan` when it should list some of those plans. It would be nice if we could render the other execution plans on the same page and just mark the broken ones somehow. Is it out of scope of this PR?

iNecas:
Good point, I agree we should hande it there as well, will work on that.

iNecas:
@adamruzicka I've added `ExecutionPlan::InvalidExecutionPlan` object, that is being used when there are issue loading all the data: this can be used still to show the data in console

adamruzicka:
Looks good to me, APT

PR #230, opened 2017-05-15
Pin Sinatra to 1.4.8

zjhuntin:


iNecas:
What is the reason for this change? The tests, as well as pure Dynflow seems to work with sinatra 2.0 just fine

iNecas:
I see some failures on rack-mock with `NoMethodError: undefined method `b' for "GET":String`, which seem to be related to https://github.com/rack/rack/commit/c7ee52f6ba4684ce917b97c1ff4c7c523760a9d2 but TBH we should just drop 1.9.3 at this point, as there is no use-case for 1.9.3 for us anymore

iNecas:
The 1.9.3 is being dropped in this pr https://github.com/Dynflow/dynflow/pull/231

adamruzicka:
So, is this PR needed not that #231 got in?

iNecas:
It seems there is an issue with ruby `2.0.0`, although for some reason, the travis is not pulling sinatra 2.0 even on newer Ruby versions, anyway, the proper fix should be to update the depndency only in Gemfile, and conditional on RUBY_VERSION '< 2.2.0'

zjhuntin:
@iNecas Should I just go ahead and close this PR in favor of something else?

iNecas:
Right now, I'm trying to figure out, why the travis is not pulling the Sinatra 2.0 in any Ruby version

iNecas:
I was not able to find the exact reason for travis to be passing, other than that it stopped passing when I removed `gem update bundler` from the test script. One would think that new version of bundler was the thing, but I was able to get the same error. Anyway, the only safe way seemed to pin the sinatra version, so doing it this way https://github.com/Dynflow/dynflow/pull/235

Issue #215, opened 2017-02-19
Compatibility with algebrick 0.7.4


mmoll:
Debian/strech will bring algebrick 0.7.4: https://packages.debian.org/ruby-algebrick

As a consequence (for foreman-proxy plugins, which use the ruby-dynflow package, which in turn uses the gem in that ruby-algebnrick package) either a version of dynflow that's compatible is needed or an updated version of algebrick, which would then need to be placed into Foreman's APT repos.

iNecas:
Thanks for letting me know, I will try to resolve this issue this week.

mmoll:
what's the status here?

iNecas:
It took a while but I've finally found the issue we had with the update of algebrick. This https://github.com/Dynflow/dynflow/pull/231 should resolve thie issue once and for all.

adamruzicka:
Closing this issue since #231 was merged

PR #231, opened 2017-05-15
Fix compatibility with algebrick 0.7.4

iNecas:
Algebrick 0.7.4 defines eql? and hash function on the objects. The
problem is that in clock, the receiver is part of the algebrick object,
which leads to the receiver to be included in the hash calculation. What
can happen is the application of the timer actually modifies the
receiver so that it alters the hash value of the object, and it failed
to delete the object from list of timers, leading to endless loops.

The most simple example is this:

   clock.ping([], 0.1, :pong, :<<)
   sleep 0.2

Once the ping realizes, the receiver changes from `[]` to `[:pong]` and
it fails to delete the timer from the list.

Fallbacking to using `object_id` as the hash function should work for us
more safely for this particular use-case.


iNecas:
Fixes #215

iNecas:
I'm also dropping 1.9.3, as the tests are failing there due to issues in some dependencies

adamruzicka:
Looks good and works as expected, merging

adamruzicka:
Thanks @iNecas !

PR #233, opened 2017-05-19
pin concurrent-ruby-edge to any version of the 0.2.x series

evgeni:
in e9c89e7 we pinned concurrent-ruby-edge to ~> 0.2.3, because that was the latest available version at that time. but in reality we really want *any* version of the 0.2.x series, just not 0.3.x. let's soften the 
dependency accordingly. this allows the use of dynflow with older c-r-e releases.

I did run the unit-tests on both, c-r-e 0.2.0 and 0.2.4, and they run fine.


evgeni:
the travis failures look unrelated?

iNecas:
Works for me

PR #234, opened 2017-05-19
DNM: see what's in env

iNecas:
To find the reason why sinatra 2.0 is not being pulled


iNecas:
I still don't understand the difference between travis env and my dev env, so that the travis handles the dependencies with taking ruby_version into account, while my env not. Anyway, long story short https://github.com/Dynflow/dynflow/pull/235 and closing this PR

PR #236, opened 2017-05-29
Port memory profiling from foreman-tasks

dLobatog:
Needs to be manual tested and tests from https://github.com/theforeman/foreman-tasks/commit/d3051b91d63a1c809b78fbd48187ba364c391a89 have to be ported, this is preliminary

### DIFF
--- a/dynflow.gemspec
+++ b/dynflow.gemspec
@@ -33,4 +33,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'activejob', '< 5.0.0'
   s.add_development_dependency "sqlite3"
   s.add_development_dependency "sinatra"
+  s.add_development_dependency 'mocha'
 end
--- a/lib/dynflow/rails/daemon.rb
+++ b/lib/dynflow/rails/daemon.rb
@@ -1,18 +1,55 @@
 require 'fileutils'
+require 'get_process_mem'
+require 'dynflow/watchers/memory_consumption_watcher'
+require 'active_support/core_ext/numeric/bytes'
 
 module Dynflow
   class Rails
     class Daemon
+      attr_reader :dynflow_memory_watcher_class, :daemons_class
+
+      # make Daemon dependency injection ready for testing purposes
+      def initialize(
+        dynflow_memory_watcher_class = ::Dynflow::Watchers::MemoryConsumptionWatcher,
+        non_default_daemons_class = nil
+      )
+        @dynflow_memory_watcher_class = dynflow_memory_watcher_class
+        @daemons_class = non_default_daemons_class
+      end
+
+      def daemons_class
+        @daemons_class || ::Daemons
+      end
+
       # Load the Rails environment and initialize the executor in this thread.
-      def run(rails_root = Dir.pwd)
+      def run(rails_root = Dir.pwd, options = {})
         STDOUT.puts('Starting Rails environment')
         rails_env_file = File.expand_path('./config/environment.rb', rails_root)
         unless File.exist?(rails_env_file)
           raise "#{rails_root} doesn't seem to be a Rails root directory"
         end
+
+        STDERR.puts("Starting dynflow with the following options: #{options}")
+
         ::Rails.application.dynflow.executor!
+
+        if options[:memory_limit] && options[:memory_limit].to_i > 0
+          ::Rails.application.dynflow.config.on_init do |world|
+            memory_watcher = initialize_memory_watcher(world, options[:memory_limit], options)
+            world.terminated.on_completion do
+              STDOUT.puts("World has been terminated")
+              memory_watcher = nil # the object can be disposed
+            end
+          end
+        end
+
         require rails_env_file
-        STDOUT.puts('Everything ready')
+        world_id = if ::Rails.application.dynflow.initialized?
+                     ::Rails.application.dynflow.world.id
+                   else
+                     nil
+                   end
+        STDOUT.puts("Everything ready for world: #{world_id}")
         sleep
       ensure
         STDOUT.puts('Exiting')
@@ -20,13 +57,6 @@ module Dynflow
 
       # run the executor as a daemon
       def run_background(command = 'start', options = {})
-        default_options = { rails_root: Dir.pwd,
-                            process_name: 'dynflow_executor',
-                            pid_dir: File.join(::Rails.root, 'tmp', 'pids'),
-                            log_dir: File.join(::Rails.root, 'log'),
-                            wait_attempts: 300,
-                            wait_sleep: 1,
-                            executors_count: (ENV['EXECUTORS_COUNT'] || 1).to_i }
         options = default_options.merge(options)
         FileUtils.mkdir_p(options[:pid_dir])
         begin
@@ -42,23 +72,19 @@ module Dynflow
         STDOUT.puts("Dynflow Executor: #{command} in progress")
 
         options[:executors_count].times do
-          Daemons.run_proc(options[:process_name],
-                           :multiple => true,
-                           :dir => options[:pid_dir],
-                           :log_dir => options[:log_dir],
-                           :dir_mode => :normal,
-                           :monitor => true,
-                           :log_output => true,
-                           :ARGV => [command]) do |*_args|
-                             begin
-                               ::Logging.reopen
-                               run(options[:rails_root])
-                             rescue => e
-                               STDERR.puts e.message
-                               ::Rails.logger.exception('Failed running Dynflow daemon', e)
-                               exit 1
-                             end
-                           end
+          daemons_class.run_proc(
+            options[:process_name],
+            daemons_options(command, options)
+          ) do |*_args|
+            begin
+              ::Logging.reopen
+              run(options[:rails_root], options)
+            rescue => e
+              STDERR.puts e.message
+              ::Rails.logger.exception('Failed running Dynflow daemon', e)
+              exit 1
+            end
+          end
         end
       end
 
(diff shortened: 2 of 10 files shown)


lib/dynflow/rails/daemon.rb
```diff
@@ -1,32 +1,54 @@
...
         unless File.exist?(rails_env_file)
           raise "#{rails_root} doesn't seem to be a Rails root directory"
         end
+
+        STDERR.puts("Starting dynflow with the following options: #{options}")
+
         ::Rails.application.dynflow.executor!
+
+        if options[:memory_limit] > 0
+          ::Rails.application.dynflow.config.on_init do |world|
```
iNecas: Ideally, this should not be dependent on rails environment

lib/dynflow/rails/daemon.rb
```diff
@@ -1,32 +1,54 @@
...
 
 module Dynflow
   class Rails
     class Daemon
+      attr_reader :dynflow_memory_watcher_class, :daemons_class
+
+      # make Daemon dependency injection ready for testing purposes
+      def initialize(
+        dynflow_memory_watcher_class = ::Dynflow::Watchers::MemoryConsumptionWatcher,
+        daemons_class = ::Daemons
```
ShimShtein: We should set to nil here, and create a default inside the `daemons_class` accessor. This way it would be accessed only on demand (from `run_background`).

lib/dynflow/rails/daemon.rb
dLobatog: It's dependent on -tasks already - https://github.com/theforeman/foreman-tasks/commit/d3051b91d63a1c809b78fbd48187ba364c391a89#diff-1f87f95c00892140947fc8f17d3f8980R34 , this module just tries to find the world used by the Rails application

dLobatog:
Fixed the problem with memory_limit, tested with https://github.com/theforeman/foreman-tasks/pull/235 and https://github.com/theforeman/foreman/pull/4316 - I was able to schedule jobs and enqueue them with `SampleJob.perform_later("test")` and `SampleJob.set(:wait => 3.seconds)` , both using bin/dynflow-executor running the executor along the Rails process.

lib/dynflow/rails/daemon.rb
iNecas: Oh I see, nm then

dynflow.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency "concurrent-ruby", '~> 1.0'
   s.add_dependency "concurrent-ruby-edge", '~> 0.2.0'
   s.add_dependency "sequel", '>= 4.0.0'
+  s.add_dependency "daemons"
```
adamruzicka: Is it necessary to make it a hard dependency of dynflow? Wouldn't it be better to put it into a group in Gemfile and make it optional?

lib/dynflow/rails/daemon.rb
```diff
@@ -42,23 +64,19 @@ def run_background(command = 'start', options = {})
...
-                               exit 1
-                             end
-                           end
+          daemons_class.run_proc(
+            options[:process_name],
+            daemons_options(command, options)
+          ) do |*_args|
+            begin
+              ::Logging.reopen
+              run(options[:rails_root], options)
```
adamruzicka: This should probably be `options[:foreman_root]`

Also when trying to run `bundle exec ruby bin/dynflow-executor run` I'm getting
```
Dynflow Executor: run in progress
Starting Rails environment
Starting dynflow with the following options: {:foreman_root=>"/home/aruzicka/git/foreman", :process_name=>"dynflow_executor", :pid_dir=>"/home/aruzicka/git/foreman/tmp/pids", :log_dir=>"/home/aruzicka/git/foreman/log", :wait_attempts=>300, :wait_sleep=>1, :executors_count=>1, :memory_limit=>0, :memory_init_delay=>7200, :memory_polling_interval=>60}
Exiting
undefined method `[]' for nil:NilClass
/home/aruzicka/git/dynflow/lib/dynflow/rails/daemon.rb:76:in `rescue in block (2 levels) in run_background': undefined method `exception' for nil:NilClass (NoMethodError)
        from /home/aruzicka/git/dynflow/lib/dynflow/rails/daemon.rb:71:in `block (2 levels) in run_background'
        from /home/aruzicka/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/daemons-1.2.4/lib/daemons/application.rb:266:in `block in start_proc'
        from /home/aruzicka/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/daemons-1.2.4/lib/daemons/application.rb:275:in `start_proc'
        from /home/aruzicka/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/daemons-1.2.4/lib/daemons/application.rb:296:in `start'
        from /home/aruzicka/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/daemons-1.2.4/lib/daemons/controller.rb:59:in `run'
        from /home/aruzicka/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/daemons-1.2.4/lib/daemons.rb:197:in `block in run_proc'
        from /home/aruzicka/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/daemons-1.2.4/lib/daemons/cmdline.rb:92:in `catch_exceptions'
        from /home/aruzicka/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/daemons-1.2.4/lib/daemons.rb:196:in `run_proc'
        from /home/aruzicka/git/dynflow/lib/dynflow/rails/daemon.rb:67:in `block in run_background'
        from /home/aruzicka/git/dynflow/lib/dynflow/rails/daemon.rb:66:in `times'
        from /home/aruzicka/git/dynflow/lib/dynflow/rails/daemon.rb:66:in `run_background'
        from bin/dynflow-executor:70:in `<main>'
```

adamruzicka: Oh, `dynflow-executor` and `foreman-tasks` each pass different options, this needs to be unified either here or in the scripts

test/daemon_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    before do
+      @dynflow_memory_watcher = Minitest::Mock.new('memory_watcher')
+      @daemons = Minitest::Mock.new('daemons')
+      @daemon = Dynflow::Rails::Daemon.new(
+        @dynflow_memory_watcher,
+        @daemons
+      )
+      @world_class = Minitest::Mock.new('dummy world factory')
+      @dummy_world = Dynflow::Testing::DummyWorld.new
+      @dummy_world.stub(:auto_execute) { nil }
```
adamruzicka: The tests will need some attention, please note Dynflow uses "plain" `minitest` instead of `mocha` which does stubbing a bit differently http://docs.seattlerb.org/minitest/Object.html#method-i-stub

lib/dynflow/rails/daemon.rb
```diff
@@ -42,23 +71,20 @@ def run_background(command = 'start', options = {})
...
-                               exit 1
-                             end
-                           end
+          daemons_class.run_proc(
+            options[:process_name],
+            daemons_options(command, options)
+          ) do |*_args|
+            begin
+              ::Logging.reopen
+              puts options[:rails_root]
```
adamruzicka: Is this a debugging leftover?

dLobatog: Yup

adamruzicka:
Took care of the failing tests, it's all green now. Full review will follow later today

iNecas:
I'm ok with intoroducing mocha for testing, if that's the case it's complicated to move the memory profiling over here, but I would definitely welcome moving everything here, as opposed to having the tests cross projects.

lib/dynflow/rails/daemon.rb
```diff
@@ -1,32 +1,61 @@
...
         rails_env_file = File.expand_path('./config/environment.rb', rails_root)
         unless File.exist?(rails_env_file)
           raise "#{rails_root} doesn't seem to be a Rails root directory"
         end
+
+        STDERR.puts("Starting dynflow with the following options: #{options}")
+
         ::Rails.application.dynflow.executor!
+
+        if !options[:memory_limit].nil? && options[:memory_limit] > 0
```
adamruzicka: Could you please pull in and adapt the tests from https://github.com/theforeman/foreman-tasks/pull/268/files? Just so we don't lose the tests once this gets in.

lib/dynflow/testing/dummy_world.rb
```diff
@@ -38,6 +38,8 @@ def persistence
         nil
       end
 
+      def auto_execute; end
+      def terminated; end
```
adamruzicka: Why are these two needed (just curious)?

adamruzicka:
Left some small comments inline, however I'm not really fond of splitting the tests

dLobatog:
@adamruzicka @iNecas Working on migrating tests - for the moment on a different branch (https://github.com/dlobatog/dynflow/tree/mem-profiling-rails) - it highlighted a few dependencies we have on Foreman with this code (`to_gb`, `::Logging` from the logging gem, etc..). The biggest issue isn't mocha, but simulating a whole `::Rails.application` object and environment when it doesn't exist.

So far I made 1/4 tests pass, so there's hope 😏 !

test/daemon_test.rb
```diff
@@ -0,0 +1,98 @@
...
+    ::Rails.stubs(:application).returns(OpenStruct.new(:dynflow => @dynflow))
+    ::Rails.stubs(:root).returns('support/rails')
+    ::Rails.stubs(:logger).returns(Logging.logger(STDOUT))
+    @dynflow.require!
+    @daemon.stubs(:sleep).returns(true) # don't pause the execution
+    @current_folder = File.expand_path('../support/rails/', __FILE__)
+    ::ActiveRecord::Base.configurations = { 'development' => {} }
+    ::Dynflow::Rails::Configuration.any_instance.stubs(:initialize_persistence).
+      returns(WorldFactory.persistence_adapter)
+  end
```
iNecas: It seems you're missing 

```ruby
def teardown
  @dynflow.world.terminate.wait
end
```

We check at every test, if there are some unset futures to protect ourselves from leaking.

dLobatog: ta, 0 errors but since it returns non-zero it fails 👍

dLobatog: Hmm.. this doesn't work because it's a DummyWorld which doesn't contain the terminate method, any ideas?

test/daemon_test.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  teardown do
+    @event.complete
```
dLobatog: Had to make sure there were no zombies by the end of testing in order to return 0

iNecas:
:+1: from me. @adamruzicka feel free to merge if you're ok with it

adamruzicka:
Merged, thanks @dLobatog !

dLobatog:
Awesome, thanks @adamruzicka ! Can we have a new release to update the dynflow version in -tasks and Foreman?

PR #240, opened 2017-07-13
Fixes #20299 - Add support for abort (force cancel)

adamruzicka:


iNecas:
Looks ok code-wise, please add some tests. I will proceed with testing in connection with foreman-tasks and rex

PR #241, opened 2017-07-19
Refs #20352 - Backup execution plans on their deletion

mbacovsky:
When execution plan is deleted it is exported along with actions and
steps. The backup is controlled by two configuration options -
backup_deleted_plans and backup_dir. The data are exported in CSV format
one file per plans, actions and steps. The files are created under the
backup_dir and daily subdirs for easier rotations.

### DIFF
--- a/lib/dynflow/config.rb
+++ b/lib/dynflow/config.rb
@@ -115,6 +115,14 @@ module Dynflow
       { 'hostname' => Socket.gethostname, 'pid' => Process.pid }
     end
 
+    config_attr :backup_deleted_plans, Algebrick::Types::Boolean do
+      false
+    end
+
+    config_attr :backup_dir, String, NilClass do
+      './backup'
+    end
+
     def validate(config_for_world)
       if defined? ::ActiveRecord::Base
         ar_pool_size = ::ActiveRecord::Base.connection_pool.instance_variable_get(:@size)
--- a/lib/dynflow/persistence.rb
+++ b/lib/dynflow/persistence.rb
@@ -40,13 +40,13 @@ module Dynflow
       end
     end
 
-    # - presunout current backup dir do persistence
-    # - v persistence se urci dir a bude ho mozne pretizit pomoci parametru
-    # - v tasks nactu z persistence current dir a pouziju ho
-    # - export_to_csv presunout do class method adapteru sequel. Volat ho z tasks
+    def delete_execution_plans(filters, batch_size = 1000, enforce_backup_dir = nil)
+      backup_dir = enforce_backup_dir || (@backup_deleted_plans ? current_backup_dir : nil)
+      adapter.delete_execution_plans(filters, batch_size, backup_dir)
+    end
 
-    def delete_execution_plans(filters, batch_size = 1000)
-      adapter.delete_execution_plans(filters, batch_size, @backup_deleted_plans ? @backup_dir : nil)
+    def current_backup_dir
+      File.join(@backup_dir, Date.today.strftime('%Y%m%d'))
     end
 
     def load_execution_plan(id)
--- a/lib/dynflow/persistence.rb
+++ b/lib/dynflow/persistence.rb
@@ -8,10 +8,12 @@ module Dynflow
 
     attr_reader :adapter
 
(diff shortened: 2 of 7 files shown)


lib/dynflow/config.rb
```diff
@@ -115,6 +115,14 @@ def method_missing(name)
       { 'hostname' => Socket.gethostname, 'pid' => Process.pid }
     end
 
+    config_attr :backup_deleted_plans, Algebrick::Types::Boolean do
```
adamruzicka: Do we even need this? Considering the `backup_dir` can be either a string or nil, thus functioning as a toggle as well

lib/dynflow/persistence.rb
```diff
@@ -38,8 +40,13 @@ def find_execution_plans(options)
       end
     end
 
+    # - presunout current backup dir do persistence
+    # - v persistence se urci dir a bude ho mozne pretizit pomoci parametru
+    # - v tasks nactu z persistence current dir a pouziju ho
+    # - export_to_csv presunout do class method adapteru sequel. Volat ho z tasks
```
adamruzicka: These probably need to go :)

lib/dynflow/persistence_adapters/sequel.rb
```diff
@@ -58,20 +60,48 @@ def find_execution_plans(options = {})
...
       end
 
-      def delete_execution_plans(filters, batch_size = 1000)
+      def delete_execution_plans(filters, batch_size = 1000, backup_dir = nil)
         count = 0
+        current_dir = current_backup_dir(backup_dir) if backup_dir
         filter(:execution_plan, table(:execution_plan), filters).each_slice(batch_size) do |plans|
           uuids = plans.map { |p| p.fetch(:uuid) }
           @db.transaction do
             table(:delayed).where(execution_plan_uuid: uuids).delete
```
adamruzicka: Why no backup for delayed plans?

mbacovsky: The backup is done mainly for later debugging and the schedules are usually not relevant to the action run. The real execution times are stored within the plans which is sufficient.

lib/dynflow/persistence.rb
mbacovsky: !

lib/dynflow/config.rb
mbacovsky: I believe not. `backup_dir` nil mean use the default dir. Option for backup on/off seems reasonable. I'll recheck when the foreman-tasks part is added

lib/dynflow/persistence.rb
```diff
@@ -40,13 +40,13 @@ def find_execution_plans(options)
...
-    # - export_to_csv presunout do class method adapteru sequel. Volat ho z tasks
+    def delete_execution_plans(filters, batch_size = 1000, enforce_backup_dir = nil)
+      backup_dir = enforce_backup_dir || (@backup_deleted_plans ? current_backup_dir : nil)
+      adapter.delete_execution_plans(filters, batch_size, backup_dir)
+    end
 
-    def delete_execution_plans(filters, batch_size = 1000)
-      adapter.delete_execution_plans(filters, batch_size, @backup_deleted_plans ? @backup_dir : nil)
+    def current_backup_dir
+      File.join(@backup_dir, Date.today.strftime('%Y%m%d'))
```
adamruzicka: Maybe some caching would be nice. Just to make sure that backups started minute before midnight won't get separated into two directories

mbacovsky: Such situation should be covered by possibility to enforce the backup dir from the caller of the delete. When we delete e.g. foreman tasks the current dir is computed in the beginning and and all the backups are called with the same dir. I'm not sure if the caching would work as the daily expiration is tricky. Any idea on solution?

mbacovsky:
A few minor changes was done. The counterpart in the foreman-tasks was sent [1] and should be ready for testing.

[1] https://github.com/theforeman/foreman-tasks/pull/271

adamruzicka:
Merged, thanks @mbacovsky !

PR #242, opened 2017-07-21
Fixes #20380 - fix argument error on :handle_peristence_error

iNecas:
The way we passed the arguments to the handle_persistence_error was
wrong before.


adamruzicka:
APT

PR #243, opened 2017-07-24
Polling sub tasks

adamruzicka:
- [X] resuming
- [x] tests
  - [X] tests for polling
  - [x] tests for polling + bulk
  - [x] tests for resume

### DIFF
--- a/lib/dynflow/action/with_polling_sub_plans.rb
+++ b/lib/dynflow/action/with_polling_sub_plans.rb
@@ -31,6 +31,22 @@ module Dynflow
       poll
     end
 
+    def resume
+      if sub_plans.all? { |sub_plan| sub_plan.error_in_plan? }
+        # We're starting over and need to reset the counts
+        %w(total failed pending success).each { |key| output.delete("#{key}_count".to_sym) }
+        initiate
+      else
+        if self.is_a?(::Dynflow::Actions::WithBulkSubPlans) && can_spawn_next_batch?
+          spawn_plans
+          suspend
+        else
+          poll
+        end
+      end
+    end
+
+
     def notify_on_finish(_sub_plans)
       suspend_and_ping
     end
--- a/lib/dynflow/action/with_polling_sub_plans.rb
+++ b/lib/dynflow/action/with_polling_sub_plans.rb
@@ -42,9 +42,11 @@ module Dynflow
     end
 
     def recalculate_counts
-      total   = sub_plans.count
-      failed  = sub_plans('state' => 'stopped', 'result' => 'error').count
-      success = sub_plans('state' => 'stopped', 'result' => 'success').count
+      total      = sub_plans.count
+      @sub_plans = nil # TODO:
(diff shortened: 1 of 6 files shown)


lib/dynflow/action/with_polling_sub_plans.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    def poll
+      recalculate_counts
+      try_to_finish || suspend_and_ping
+    end
+
+    def wait_for_sub_plans(sub_plans)
+      increase_counts(sub_plans.count, 0)
+      if is_a?(::Dynflow::Action::WithBulkSubPlans)
```
iNecas: Why do we need this check?

adamruzicka: We don't really need it, but it makes sense to start polling after all the sub plans were spawned.

lib/dynflow/action/with_polling_sub_plans.rb
```diff
@@ -0,0 +1,72 @@
...
+      suspend { |suspended_action| ping suspended_action }
+    end
+
+    def ping(suspended_action)
+      world.clock.ping suspended_action, REFRESH_INTERVAL, Poll
+    end
+
+    def recalculate_counts
+      total      = sub_plans.count
+      @sub_plans = nil # TODO:
```
iNecas: Planning to clean this TODOs, right? :)

test/action_test.rb
```diff
@@ -442,6 +481,34 @@ def run(event = nil)
           resumed_plan.state.must_equal :stopped
           resumed_plan.result.must_equal :success
         end
+
+        describe ::Dynflow::Action::WithPollingSubPlans do
+          include TestHelpers
+
+          let(:klok) { Dynflow::Testing::ManagedClock.new }
```
iNecas: Any reason not to call it `:clock`?

iNecas:
Codewise looks good, waiting for the test of the tests to start testing this manually

test/action_test.rb
adamruzicka: not really, will change it

lib/dynflow/action/with_polling_sub_plans.rb
iNecas: If they are dependent, having two commits in one PR, with the final set of changes would be the best.

adamruzicka: Both PRs should now containing commit https://github.com/Dynflow/dynflow/pull/243/commits/fdf57ed3273f141a4bf56687072f1b3eddeb6ca5 with the common changes

iNecas:
Tested and works well. Thanks @adamruzicka. Please crate an issue for foreman_remote_execution to start using it and let's add it there. I think it's a prerequisity for http://projects.theforeman.org/issues/19937/ to work as expected (from the user point of view)

PR #244, opened 2017-08-17
Execution plan cleaner

adamruzicka:
It is disabled by default, needs to be enabled in a world's config. For the usage on smart-proxy this should be good to go since there's mostly one or two kinds of tasks (rex and ansible), so we probably don't need fine grained controls for setting different interval for different actions.

### DIFF
--- a/lib/dynflow/actors/execution_plan_cleaner.rb
+++ b/lib/dynflow/actors/execution_plan_cleaner.rb
@@ -1,0 +1,52 @@
+module Dynflow
+  module Actors
+    class ExecutionPlanCleaner
+      attr_reader :core
+
+      def initialize(world, options = {})
+        @world = world
+        @options = options
+      end
+
+      def core_class
+        Core
+      end
+
+      def spawn
+        Concurrent.future.tap do |initialized|
+          @core = core_class.spawn(:name => 'execution-plan-cleaner',
+                                   :args => [@world, @options],
+                                   :initialized => initialized)
+        end
+      end
+
+      def clean!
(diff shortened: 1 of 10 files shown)


lib/dynflow/actors/execution_plan_cleaner.rb
```diff
@@ -0,0 +1,52 @@
...
+        end
+
+        def start
+          set_clock
+          clean!
+        end
+
+        def clean!
+          plans = @world.persistence.find_old_execution_plans(Time.now.utc - @max_age)
+          @world.persistence.delete_execution_plans(uuid: plans.map(&:id))
```
iNecas: Doesn't this have the issue with the blank filer on `uuid => []` matching all the plans? I guess we should probably just fix this case directly in sequel persistence adapter, to avoid this issues in once and for all, as I don't see where it would actually be useful.

adamruzicka: Just tried it, looks it is safe to use now

adamruzicka:
Added  some tests, let's see how Mr. T likes them

adamruzicka:
The tests are green

lib/dynflow/actors/execution_plan_cleaner.rb
iNecas: Ah, sry, my bad, I was trying it against find_execution_paln with wrong filter

iNecas:
Tests speak for themselves: thanks @adamruzicka. Merging now

PR #245, opened 2017-08-17
Dead letter handling

adamruzicka:
Spawns one dead letter handler per world, dead letters can be matched by sender, receiver and the message. Matching can be either of `::Dynflow::DeadLetterHandler::Any` (always match), `Proc` which is evaluated with the object or anything else which is checked using `==` against the object.

### DIFF
--- a/lib/dynflow.rb
+++ b/lib/dynflow.rb
@@ -23,6 +23,7 @@ module Dynflow
 
   require 'dynflow/utils'
   require 'dynflow/round_robin'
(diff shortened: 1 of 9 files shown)


iNecas:
I have to test but looks good codewise

iNecas:
Rubocop is not happy though

adamruzicka:
It needed some manual intervention but we're green now

iNecas:
Tests are happy, I am happy :)

PR #246, opened 2017-08-18
Fixes #20644 - remove google-code-prettify

iNecas:
It's been used back in the days, when json was used for showing
the actions input/output. We switched to yaml in the meantime,
therefore no need for this code anymore.

Removed javascript is good javascript.


adamruzicka:
Merged, thanks @iNecas !

PR #247, opened 2017-08-18
Rescue strategy fixes and enhancements

iNecas:
This PR includes two fixes for rescue strategy

Fixes #20647 - fix calculation of rescue strategy for complex actions

Since ba6300f, we were considering also pending and successful steps when
calculating the rescue strategy. This made the auto_rescue not working
properly, as any action with `Pause` rescue strategy (which is considered
by default) would lead to pause of the task.

This change makes the rescue strategy working properly again, while keeping
the Fail strategy working as well.

Fixes #20648 - support for multiple auto-rescues in the plan

Currently, when an error occurs and skip rescue strategy applies, it
auto-rescues only at the first failure: the auto-rescue is not applied for
further failures.

After this change, we don't check only if the plan was already rescued or
not, but we check, if there is some failed action that we have not tried to
rescue from yet: this is better check to ensure we don't run into endless
loop, while allowing to auto-rescue several times in a row.

### DIFF
--- a/lib/dynflow/director.rb
+++ b/lib/dynflow/director.rb
@@ -80,7 +80,7 @@ module Dynflow
       @world = world
       @logger = world.logger
       @execution_plan_managers = {}
-      @plan_ids_in_rescue = Set.new
+      @rescued_steps = {}
     end
 
     def start_execution(execution_plan_id, finished)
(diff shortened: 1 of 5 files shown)


lib/dynflow/director.rb
```diff
@@ -146,14 +146,20 @@ def finish_manager(manager)
...
       return false if @world.terminating?
-      @world.auto_rescue && manager.execution_plan.state == :paused &&
-        !@plan_ids_in_rescue.include?(manager.execution_plan.id)
+      return false unless @world.auto_rescue && manager.execution_plan.state == :paused
+      # we have not rescued this plan yet
+      return true unless @rescued_steps.key?(manager.execution_plan.id)
+      # we have rescued this plan already, but a different step has failed now
+      # we do this check to prevent endless loop, if we always failed on the same steps
+      failed_step_ids = manager.execution_plan.failed_steps.map(&:id).to_set
+      return true unless (failed_step_ids - @rescued_steps[manager.execution_plan.id]).empty?
```
adamruzicka: Wouldn't this be more readable rewritten as a if-elsif-else? I'm not really all that fond of multiple early returns.
```ruby
if @world.terminating? || !(@world.auto_rescue && manager.execution_plan.state == :paused)
  false
elsif !@rescued_steps.key?(manager.execution_plan.id)
  # we have not rescued this plan yet
  true
else
  # we have rescued this plan already, but a different step has failed now
  # we do this check to prevent endless loop, if we always failed on the same steps
  failed_step_ids = manager.execution_plan.failed_steps.map(&:id).to_set
  !(failed_step_ids - @rescued_steps[manager.execution_plan.id]).empty?
end
```

iNecas: Good point: it indeed looks better. Updated.

iNecas:
Thanks @adamruzicka for testing and review

PR #248, opened 2017-08-21
Make execution plan cleaner log its activity

adamruzicka:


iNecas:

PR #251, opened 2017-09-13
Fixes #20809 - Add interface for retrieving pending step counts

adamruzicka:


lib/dynflow/web/console.rb
```diff
@@ -37,6 +37,15 @@ class Console < Sinatra::Base
         erb :worlds
       end
 
+      post('/worlds/pending_steps') do
+        @worlds = world.coordinator.find_worlds
+        @worlds.each do |w|
+          steps = world.get_pending_steps(w.data['id'], nil, 5).value!.values.sum
```
iNecas: I'm getting ```NoMethodError at /worlds/pending_steps undefined method `sum' for []:Array``` when running pure dynflow, without Rails, as this methods comes from ActiveSupport

lib/dynflow/dispatcher.rb
```diff
@@ -15,14 +15,20 @@ module Dispatcher
...
+      variants Event, Execution, Ping, Steps
     end
 
     Response = Algebrick.type do
       variants Accepted = atom,
                Failed   = type { fields! error: String },
                Done     = atom,
-               Pong     = atom
+               Pong     = atom,
+               PendingSteps = type { fields! pending_steps: Hash }
```
iNecas: I'm not sure if `Steps` is the best word, as it has already some meaning: the items there might be either steps or events for the steps: it would be more like `execution items`

lib/dynflow/web/console.rb
adamruzicka: Actually it was added to pure ruby in 2.4.0, will fix anyway.

lib/dynflow/dispatcher.rb
adamruzicka: True, I'll try to come up with better name for it

adamruzicka:
API still pending

adamruzicka:
Created [separate issue](http://projects.theforeman.org/issues/21093) for the API part

lib/dynflow/executors/parallel/pool.rb
```diff
@@ -62,6 +71,12 @@ def start_termination(*args)
           try_to_terminate
         end
 
+        def execution_items(execution_plan_id = nil)
```
iNecas: Since we added more things than just the execution items. something more generic, such as `execution_status` or something like that would match more what it actually does

lib/dynflow/web/console.rb
```diff
@@ -37,6 +37,16 @@ class Console < Sinatra::Base
         erb :worlds
       end
 
+      post('/worlds/execution_items') do
```
iNecas: Again: Since we added more things than just the execution items. something more generic, such as execution_status or something like that would match more what it actually does.

lib/dynflow/web/console.rb
```diff
@@ -37,6 +37,16 @@ class Console < Sinatra::Base
...
+        @worlds = world.coordinator.find_worlds
```
iNecas: We should limit this to active executors only (so `find_worlds(true)`) as no reason for reaching to the client worlds.

iNecas:
I don't want the PR to grow to much, there fore here are some post-merge ideas to continue:

  * separate the execution worlds table from the client worlds. The client worlds can stay as table, but the executor ones might deserve something like patternfly cards. Btw. I wonder how much work it would be to teach dynflow some patternfly :)

* one reason why the table might not be best for the executor worlds is this idea: showing the info
about steps (including event info and link to the execution plan) that are being executed (ideally with
info on how long the item has been processed: this would allow us to easily see, what work items
are blocking the executors).

Again: I'm ok with taking these outside of the PR, but would be great addition to the console regarding
showing the info about the runtime

adamruzicka:
Those items would be definitely nice to have but will have to wait for another PR

lib/dynflow/executors/parallel/pool.rb
```diff
@@ -23,6 +23,15 @@ def empty?
             @jobs.empty?
           end
 
+          def execution_status(execution_plan_id = nil)
+            source = if execution_plan_id.nil?
+                       @jobs
+                     else
+                       { execution_plan_id => @jobs.fetch(execution_plan_id, []) }
+                     end
+            source.reduce({}) { |acc, cur| acc.update(cur.first => cur.last.count) }
```
iNecas: What about using destructuring assignment here?:

```ruby
source.reduce({}) { |acc, (plan_id, work_items)| acc.update(plan_id => work_items.count) }
```

iNecas:
Tested and worked well. Thanks @adamruzicka

PR #253, opened 2017-09-28
Introduce Singleton action module

adamruzicka:
An action into which this module is included (singleton action from
now on) on which there should be only one instance running at any
given time.

This means the action tries to obtain a lock at the beginning of its
plan phase and fails if it doesn't acquire it. At each Action#run it
checks if it is still holding the lock.  If it does, it goes on,
otherwise tries to reacquire it. It behaves the same way during the
lock reacquiring as in plan phase.

Finally, it releases the lock at the end of finalize phase. If
things go wrong, locks should get released when the execution plan
holding them gets invalidated.

Please make sure singleton actions release the lock they're holding
in case they encounter any error from which they can't recover.

### DIFF
--- a/examples/singletons.rb
+++ b/examples/singletons.rb
@@ -17,12 +17,11 @@ DESC
 require_relative 'example_helper'
 
 class SingletonExample < Dynflow::Action
+  middleware.use ::Dynflow::Middleware::Common::Singleton
   include Dynflow::Action::Singleton
 
   def run
-    with_valid_singleton_lock do
-      sleep 10
-    end
+    sleep 10
   end
 end
 
--- a/examples/singletons.rb
+++ b/examples/singletons.rb
@@ -1,0 +1,55 @@
+#!/usr/bin/env ruby
+example_description = <<DESC
+
+  Sub Plans Example
+  ===================
+
+  This example shows, how singleton actions can be used for making sure
+  there is only one instance of the action running at a time.
+
+  Singleton actions try to obtain a lock at the beggining of their plan
+  phase and fail if they can't do so. In run phase they check if they
+  have the lock and try to acquire it again if they don't. These actions
+  release the lock at the end of their finalize phase.
+
+DESC
+
+require_relative 'example_helper'
+
+class SingletonExample < Dynflow::Action
+  include Dynflow::Action::Singleton
+
+  def run
+    with_valid_singleton_lock do
+      sleep 10
+    end
+  end
+end
+
+class SingletonExampleA < SingletonExample; end
+class SingletonExampleB < SingletonExample; end
+
+if $0 == __FILE__
+  ExampleHelper.world.action_logger.level = Logger::INFO
+  ExampleHelper.world
+  t1 = ExampleHelper.world.trigger(SingletonExampleA)
+  t2 = ExampleHelper.world.trigger(SingletonExampleA)
+  ExampleHelper.world.trigger(SingletonExampleA) unless SingletonExampleA.singleton_locked?(ExampleHelper.world)
+  t3 = ExampleHelper.world.trigger(SingletonExampleB)
+  db = ExampleHelper.world.persistence.adapter.db
+
+  puts example_description
+  puts <<-MSG.gsub(/^.*\|/, '')
+    |  3 execution plans were triggered:
+    |  #{t1.id} should finish successfully
+    |  #{t3.id} should finish successfully because it is a singleton of different class
+    |  #{t2.id} should fail because #{t1.id} holds the lock
+    |  
+    |  You can see the details at
+    |    http://localhost:4567/#{t1.id}
+    |    http://localhost:4567/#{t2.id}
+    |    http://localhost:4567/#{t3.id}
+    |
+  MSG
+  ExampleHelper.run_web_console
+end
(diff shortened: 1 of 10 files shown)


adamruzicka:
Rewritten katello's action with this are being worked on here https://github.com/adamruzicka/katello/tree/singleton-actions

examples/singletons.rb
```diff
@@ -0,0 +1,55 @@
...
+
+DESC
+
+require_relative 'example_helper'
+
+class SingletonExample < Dynflow::Action
+  include Dynflow::Action::Singleton
+
+  def run
+    with_valid_singleton_lock do
```
iNecas: What about encapsulating the locking calls into a middleware, so that the actions developer don't have to care/think not to forget to put the right lock/unlock calls at appropriate places?

examples/singletons.rb
```diff
@@ -0,0 +1,54 @@
...
+  have the lock and try to acquire it again if they don't. These actions
+  release the lock at the end of their finalize phase.
+
+DESC
+
+require_relative 'example_helper'
+
+class SingletonExample < Dynflow::Action
+  middleware.use ::Dynflow::Middleware::Common::Singleton
+  include Dynflow::Action::Singleton
```
iNecas: what about putting this into the `Dynflow::Actions::Singleton`:

```ruby
def self.included(base)
  base.middleware.use ::Dynflow::Middleware::Common::Singleton
end
```

So that the people don't have to think about that (and forget) when they add it to their code. Also, a corresponding PR to docs would be nice while we have that in fresh memory.

lib/dynflow/middleware/common/singleton.rb
```diff
@@ -0,0 +1,24 @@
...
+        end
+      end
+
+      def run(*args)
+        action.singleton_lock! unless action.holds_singleton_lock?
+        pass(*args)
+        action.singleton_unlock! unless action.respond_to?(:finalize)
+      end
+
+      def finalize(*args)
```
iNecas: Just capturing our yesterday's conversation: there is `finalize_phase` method, that wraps the whole finalize phase. It hoverer, doesn't expose the action. Therefore, I would vote for keeping it this way.

examples/singletons.rb
adamruzicka: Good idea, will do

lib/dynflow/middleware/common/singleton.rb
```diff
@@ -0,0 +1,24 @@
...
+        pass(*args)
+        unless action.respond_to?(:run) || action.respond_to?(:finalize)
+          action.singleton_unlock!
+        end
+      end
+
+      def run(*args)
+        action.singleton_lock! unless action.holds_singleton_lock?
+        pass(*args)
+        action.singleton_unlock! unless action.respond_to?(:finalize)
```
iNecas: Few questions:

1. what happens when the action goes to suspended state during the run phase?
2. what happens if the action fails? What if the user skips the action on failure? Would the lock stay active?
3. what happens to the lock when the user deletes the execution plan? Shouldn't we have a mechanism to invalidate the locks if the holding execution plan is missing? Also probably deleting the lock on execution plan deletion would be a good idea
4. what should happen on the task pause? Should it unlock or keep the lock. I would vote for unlocking, to let spawn a new task instead of resuming.

Wouldn't be a solution to avoid this questions to move the releasing of the `SingletonActionLock` to the execution plan, and hook it on the change of the execution plan state?

lib/dynflow/middleware/common/singleton.rb
adamruzicka: Meanwhile I somehow managed to get it working using finalize_phase. Explanatory comment incoming

adamruzicka:
Changes in 8732c3e6efa18cd7a7aa0072100deafed99868af:
- the action tries to lock itself before `#plan`
- if the action has no run or finalize steps (or failed during planning), all locks owned by this execution plan are released at the end of `plan_phase`.
- the action asserts it holds the lock before `#run`
- if the execution plan has no finalize steps, the action unlocks itself after `#run`
- if all finalize steps finished successfully, all locks owned by this execution plan are release at the end of `finalize_phase`

lib/dynflow/middleware/common/singleton.rb
adamruzicka: 1) action suspends itself and keeps the lock (we even have a test for that), when it receives an event it asserts it still holds the lock and continues
2) The execution plan goes to paused-error, and may (in case of any unhandled exception) or may not (in case `#error!` was used) keep the lock. By default it doesn't do anything with the lock when skipping.
3) the locks get released when the execution plan is invalidate. This is probably not enough and a more robust solution is needed to handle all the other cases.
4) Originally I thought the task should keep the lock until it goes to stopped, but unlocking would probably be better. Letting a new task spawn sounds good

Moving the unlocking to execution plan would be probably nicer, I'll try looking into it

adamruzicka:
Docs PR opened https://github.com/Dynflow/dynflow/pull/254

examples/singletons.rb
iNecas: so you can remove it now

lib/dynflow/execution_plan.rb
```diff
@@ -505,6 +509,10 @@ def self.steps_from_hash(step_ids, execution_plan_id, world)
       end
     end
 
+    def unlock_all_singleton_locks!
+      actions.select(&:holds_singleton_lock?).each(&:singleton_unlock!)
```
iNecas: Loading all actions every time the plan goes to paused/stopped state might be quite resource-hungry. Going directly though coordinator API to just release the singleton locks by `execution_plan-#{id}` owner would be probably much faster.

iNecas:
Thanks @adamruzicka! Tested and works well!

PR #255, opened 2017-10-13
Cache Ping messages

adamruzicka:
Some other projects use Pings quite extensively (run them at the start
of every actions) so it seemed like a good idea to cache them to reduce
the number of sent messages.

This commit introduces such mechanism. Cache records are added either
when a response to a Ping or a Ping is received. This way, both
participants in the communication have the other one in their own
caches.

Furthermore a world tries to announce its availability (pings all other
worlds) at the end of its #initialize.

### DIFF
--- a/lib/dynflow/config.rb
+++ b/lib/dynflow/config.rb
@@ -113,6 +113,10 @@ module Dynflow
       true
     end
 
+    config_attr :announce, Algebrick::Types::Boolean do
+      true
+    end
+
     config_attr :auto_validity_check, Algebrick::Types::Boolean do |world, config|
       !!config.executor
     end
--- a/lib/dynflow/config.rb
+++ b/lib/dynflow/config.rb
@@ -98,11 +98,19 @@ module Dynflow
     end
 
     config_attr :executor, Executors::Abstract, FalseClass do |world, config|
-      Executors::Parallel.new(world, config.queues)
+      Executors::Parallel.new(world, config.executor_heartbeat_interval, config.queues)
     end
 
     config_attr :executor_semaphore, Semaphores::Abstract, FalseClass do |world, config|
       Semaphores::Dummy.new
+    end
+
+    config_attr :executor_heartbeat_interval, Integer do
+      15
+    end
+
+    config_attr :ping_cache_age, Integer do
+      60
     end
 
     config_attr :connector, Connectors::Abstract do |world|
--- a/lib/dynflow/config.rb
+++ b/lib/dynflow/config.rb
@@ -113,10 +113,6 @@ module Dynflow
       true
     end
 
-    config_attr :announce, Algebrick::Types::Boolean do
-      true
-    end
-
     config_attr :auto_validity_check, Algebrick::Types::Boolean do |world, config|
       !!config.executor
     end
--- a/lib/dynflow/dispatcher/client_dispatcher.rb
+++ b/lib/dynflow/dispatcher/client_dispatcher.rb
@@ -26,16 +26,35 @@ module Dynflow
         end
       end
 
+      class PingCache
+        def initialize(age = 10)
+          @cache   = {}
+          @max_age = age
+        end
+
+        def add_record(id)
+          @cache[id] = Time.now
+        end
+
+        def fresh_record?(id)
+          @cache[id] && @cache[id] >= (Time.now - @max_age)
+        end
+      end
+
+      attr_reader :ping_cache
       def initialize(world)
         @world            = Type! world, World
         @last_id          = 0
         @tracked_requests = {}
         @terminated       = nil
+        @ping_cache       = PingCache.new
       end
 
       def publish_request(future, request, timeout)
-        track_request(future, request, timeout) do |tracked_request|
-          dispatch_request(request, @world.id, tracked_request.id)
+        with_ping_request_caching(request, future) do
+          track_request(future, request, timeout) do |tracked_request|
+            dispatch_request(request, @world.id, tracked_request.id)
+          end
         end
       end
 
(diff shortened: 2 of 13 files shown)


ShimShtein:
If I am reading the code correctly, you are saving last ping in DB, and associate it with the server record.
I think ping info should be "remembered" on client level, not server - the fact that one client can access the server does not mean that other clients will be able to do so too. Also this info could be stored in-memory, so it will remove the query to the DB.

adamruzicka:
> the fact that one client can access the server does not mean that other clients will be able to do so too

This is usually when direct connections are in play. However in dynflow world/executors communicate through one central db. So if A can reach B through the DB, the we can safely assume C can reach B through the DB if C can reach the DB.

> Also this info could be stored in-memory, so it will remove the query to the DB.

My previous implementation was like this. The downside was the information had to be kept in several copies (one for each world) and there was no easy way of exposing it in the console.

lib/dynflow/dispatcher/client_dispatcher.rb
```diff
@@ -26,16 +26,57 @@ def success!(resolve_to)
...
+        end
+      end
+
+      attr_reader :ping_cache
       def initialize(world)
         @world            = Type! world, World
         @last_id          = 0
         @tracked_requests = {}
         @terminated       = nil
+        @ping_cache       = PingCache.new world
```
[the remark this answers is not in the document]
adamruzicka: Moved it to a constant to keep things simple

lib/dynflow/dispatcher/client_dispatcher.rb
```diff
@@ -26,16 +26,84 @@ def success!(resolve_to)
...
+      # World's coordinator record include the time when was the world
+      # seen for the last time. This class can be used to query this
+      # information and determine whether the record is "fresh enough"
+      # or whether the Ping really needs to be sent.
+      class PingCache
+        # Format string used for formating and parsing times
+        TIME_FORMAT = '%Y-%m-%d %H:%M:%S.%L'.freeze
+
+        # Maximum age of a record in seconds, older records are not considered fresh
+        PING_CACHE_AGE = 10
```
iNecas: With the light of https://bugzilla.redhat.com/show_bug.cgi?id=1538688, I wonder if we could increase this to something like 70, and do some heartbeat internally to update the ping cache when the executor is live, so that we would avoid the polling completely, which might be an issue when the executor is too busy.

iNecas:
@adamruzicka with recent issues around https://bugzilla.redhat.com/show_bug.cgi?id=1538688, I would suggest reconsidering the approach and updating the heartbeat status by the server and storing it in coordinator record: the ping then would be checking this record, rather than actually trying to talk to the executor. What do you think?

adamruzicka:
Implemented executor heartbeat and longer freshness interval. Now Pings never reach executors (executors are checked only through the cache), but pings are still used the check livenes of client worlds.

lib/dynflow/dispatcher/client_dispatcher.rb
```diff
@@ -26,16 +26,95 @@ def success!(resolve_to)
...
+      # World's coordinator record include the time when was the world
+      # seen for the last time. This class can be used to query this
+      # information and determine whether the record is "fresh enough"
+      # or whether the Ping really needs to be sent.
+      class PingCache
+        # Format string used for formating and parsing times
+        TIME_FORMAT = '%Y-%m-%d %H:%M:%S.%L'.freeze
+
+        # Maximum age of a record in seconds, older records are not considered fresh
+        PING_CACHE_AGE = 60
```
iNecas: I wonder if we shouldn't expose this and the `HEARTBEAT_INTERVAL` via dynflow config?

test/test_helper.rb
```diff
@@ -96,6 +96,7 @@ def self.test_world_config
     config.backup_deleted_plans = false
     config.backup_dir           = nil
     config.queues.add(:slow, :pool_size => 1)
+    config.announce             = false
```
iNecas: I guess some residual that causes the tests to fail?

lib/dynflow/web/console.rb
```diff
@@ -50,16 +50,14 @@ class Console < Sinatra::Base
       end
 
       post('/worlds/check') do
-        load_worlds
         @validation_results = world.worlds_validity_check(params[:invalidate])
-        @worlds = world.coordinator.find_worlds
+        load_worlds
```
iNecas: Was thre any reason for this change? Just trying to understand what's the reasoning here

lib/dynflow/dispatcher/client_dispatcher.rb
```diff
@@ -80,14 +159,27 @@ def dispatch_response(envelope)
               (on ~Failed do |msg|
                  resolve_tracked_request(envelope.request_id, Dynflow::Error.new(msg.error))
                end),
-              (on Done | Pong do
+              (on Done do
+                 resolve_tracked_request(envelope.request_id)
+               end),
+              (on Pong do
+                 add_ping_cache_record(envelope.sender_id)
```
iNecas: If I read the patch correctly, we're not currently using the client pongs in cache at all. Is that intentional?

lib/dynflow/config.rb
```diff
@@ -98,13 +98,21 @@ def queues
     end
 
     config_attr :executor, Executors::Abstract, FalseClass do |world, config|
-      Executors::Parallel.new(world, config.queues)
+      Executors::Parallel.new(world, config.executor_heartbeat_interval, config.queues)
```
iNecas: This needs to be updated also here https://github.com/Dynflow/dynflow/blob/master/lib/dynflow/rails/configuration.rb#L158, otherwise the rails integration breaks.

iNecas:

PR #256, opened 2017-10-23
Fixes #21423 - Allow showing of plans without root_plan_step

adamruzicka:


web/views/show.erb
```diff
@@ -52,7 +52,7 @@
 
 <div class="tab-content">
   <div class="tab-pane" id="plan">
-    <% if @plan.valid? %>
+    <% if @plan.valid? && !@plan.root_plan_step.nil? %>
```
iNecas: I would preffer `@plan.root_plan_step` over `!@plan.root_plan_step.nil?`

adamruzicka: It indeed looks nicer

iNecas:
Thanks @adamruzicka

PR #257, opened 2017-11-02
Add support for makara adapter

iNecas:
Makara can be used as a wrapper around the standar ActiveRecord adapter.
This patch should allow users using Dynflow while using this adapter.

See [1] for such an example.

[1] - https://groups.google.com/d/msg/foreman-users/ORak_xvLi-U/Qy56C3SSAAAJ


iwalmsley:
This is great as a workaround (just need to edit the database.yml to include a hosts: entry which Sequel can pickup).  Without this builds don't even work full stop, with the inclusion of dynflow into 1.16. So if this could be merged that'd be awesome.

adamruzicka:
@iNecas, @iwalmsley: Thanks for the code and testing!

PR #258, opened 2017-11-08
Process entire clock during InThreadExecutor

jlsherrill:


lib/dynflow/testing/in_thread_executor.rb
```diff
@@ -34,7 +34,7 @@ def event(execution_plan_id, step_id, event, future = Concurrent.future)
       end
 
       def clock_tick
-        @world.clock.progress
+        @world.clock.progress_all
```
iNecas: One suggestion: the `progress_all` method could take an array of events that should be skipped (matching the `what` attribute of the event). We can state here the events that we don't care about with the InThreadExecutor (such as `:periodic_check_inbox` from https://github.com/Dynflow/dynflow/blob/master/lib/dynflow/connectors/database.rb#L77, that was causing the troubles in your case.

iNecas:
Would you mind writing some tests for it. I would suggest something like this https://github.com/Dynflow/dynflow/blob/master/test/testing_test.rb#L169 with an addition, that there would be some unrelated `clock.ping` called before the plan was triggered.

jlsherrill:
having trouble writing a test that will reproduce the issue:

```
        q = Queue.new
        q << :foo

        world.clock.ping q, 0, :a
        world.clock.ping q, 0.0001, :b
        world.clock.ping q, 0.0002, :c
        world.execute(execution_plan.id).value.tap do |plan|
          plan.state.must_equal :stopped
        end
```

Still seems to run fine (was trying something simple, before adapting to periodic_check_inbox)

adamruzicka:
@jlsherrill I think this might be the test you're looking for. I assume in the test you wrote the execution didn't depend on the clock.
```ruby
describe "PR258 reproducer" do
  class BreakingAction < ::Dynflow::Action
    def run(event = nil)
      if output[:suspended].nil?
        output[:suspended] = true
        suspend do |action|
          world.clock.ping(action, 1000, nil)
        end
      end
    end
  end

  let :world do
    WorldFactory.create_world(Dynflow::Testing::InThreadWorld)
  end

  let :execution_plan do
    world.plan(BreakingAction)
  end

  it 'breaks' do
    q = Queue.new
    20.times { |i| world.clock.ping q, 0.0002, i }
    
    f = world.execute(execution_plan.id)

    # This deadlocks the test
    f.wait
    
    f.value.tap do |plan|
      plan.state.must_equal :stopped
    end
  end
end

```

jlsherrill:
@adamruzicka that worked great!  thank you so much! 👍

lib/dynflow/testing/managed_clock.rb
```diff
@@ -26,6 +26,13 @@ def progress
         end
       end
 
+      def progress_all(ignore_list)
+        until @pending_pings.empty? || !ignore_list.include?(@pending_pings.first.what.try(:value))
```
jlsherrill: I'm sure there is some better way than calling `.try(:value)`, I initially was calling .value, but it hit an entry in @pending_pings that did not have that:  Algebrick::Types::None:Algebrick::Atom

Is there some better way to handle this condition?

adamruzicka: Well, you could always check the class of `@pending_pings.first` before checking if it is in the list, but its far from ideal. @iNecas do you recall if there's some algebrick trick for handling this situation?

test/testing_test.rb
```diff
@@ -166,6 +166,41 @@ def plan_input
...
+      end
+
+      let :world do
+        WorldFactory.create_world(Dynflow::Testing::InThreadWorld)
+      end
+
+      let :execution_plan do
+        world.plan(BreakingAction)
+      end
+
```
iNecas: I would recommand being positive (using "it works" rather than "it breaks" attitude :)

lib/dynflow/testing/managed_clock.rb
iNecas: Not a better approach I can think of: also, long term I would like to get rid of algebrick for readability reasons, so I would rather avoid some of it's advanced features

test/testing_test.rb
jlsherrill: Updated to something a little more positive :)

test/testing_test.rb
```diff
@@ -166,6 +166,41 @@ def plan_input
       end
     end
 
+    describe "in thread executor hanging test" do
```
iNecas: `describe "in thread executor hanging test"` could be `describe "in thread executor with unrelated events in clock"`

lib/dynflow/testing/managed_clock.rb
iNecas: Just be aware that `try` does not come with pure ruby and we don't have runtime dependency on active support.

jlsherrill: oh right!  I'm spoiled living in a 🛤 world

lib/dynflow/testing/managed_clock.rb
```diff
@@ -26,6 +26,18 @@ def progress
         end
       end
 
+      def skip_first?(ignored_subject)
```
adamruzicka: Maybe a name in plural would be better for the argument since we're expecting it to be an array

lib/dynflow/testing/managed_clock.rb
```diff
@@ -26,6 +26,18 @@ def progress
...
+        item = @pending_pings.first
+        item && item.what.respond_to?(:value) && ignored_subject.include?(item.what.value)
+      end
+
+      def progress_all(ignored_subject)
```
adamruzicka: Ditto

lib/dynflow/testing/managed_clock.rb
jlsherrill: good point! updated

adamruzicka:
The tests are green and it looks good to me. Anything to add, @iNecas ?

lib/dynflow/testing/managed_clock.rb
```diff
@@ -26,6 +26,18 @@ def progress
         end
       end
 
+      def skip_first?(ignored_subjects)
+        item = @pending_pings.first
+        item && item.what.respond_to?(:value) && ignored_subjects.include?(item.what.value)
+      end
+
+      def progress_all(ignored_subjects)
+        until @pending_pings.empty? || !skip_first?(ignored_subjects)
```
iNecas: I believe the `ignored_subjects` should be a parameter of progress, and in case should be ignored, just do `shift` but not actually applying it. I'm not sure this code does what it should, in terms of just dropping the events that should be ignored.

jlsherrill: it does work (it just also apply's ignored pings, which i can see maybe is not desired).  will fix

jlsherrill: updated!

lib/dynflow/testing/managed_clock.rb
```diff
@@ -17,12 +17,16 @@ def ping(who, time, with_what = nil, where = :<<)
...
 
-      def progress
+      def progress(ignored_subjects = [])
         if next_ping = @pending_pings.shift
-          # we are testing an isolated system = we can move in time
-          # without actually waiting
-          @current_time = next_ping.when
-          next_ping.apply
+          if next_ping.what.respond_to?(:value) && ignored_subjects.include?(next_ping.what.value)
+            progress(ignored_subjects)
```
iNecas: There is a risk now, that we would not ignore some ping that is not related to the specific task, and we could eventually get stuck as well. I think we should do just plain `return` here, and in `progress_all`, just do `progress(ignored_subjects) until @pending_pings.empty?`

I know I'm getting annoying here, but we're getting threre :)

jlsherrill: @iNecas updated

iNecas:
@jlsherrill 0.8.33 is all yours

PR #259, opened 2017-11-21
Fixes #21716 - Fix cancelling when using bulk actions with polling

adamruzicka:


adamruzicka:
Fixed and rebased

iNecas:
Ruboocop is not happy

adamruzicka:
Rubocop is finally happy

PR #260, opened 2017-12-11
[WIP] Performance fixes

adamruzicka:
- [x] tests for the looping bug
- [X] tests for count querying
- [x] rubocop

### DIFF
--- a/lib/dynflow/action/with_bulk_sub_plans.rb
+++ b/lib/dynflow/action/with_bulk_sub_plans.rb
@@ -92,7 +92,7 @@ module Dynflow
     end
 
     def can_spawn_next_batch?
-      total_count - output[:success_count] - output[:pending_count] - output[:failed_count] > 0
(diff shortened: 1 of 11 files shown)


iNecas:
Makes snese, pending tests

adamruzicka:
Apparently the savings from polling more efficiently are huge. Without the patch a test REX job with fake connector against 10k hosts spent ~73 minutes on polling. A job with the same parameters and the patch spent ~1 minute on polling.

```
# During the job there were ~100 polls
# Time how long a poll took in seconds without the fix
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
   2.00   21.00   44.00   44.44   67.50   92.00 

# Time how long a poll took in seconds with the fix
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
  0.000   0.000   0.000   0.604   1.000   3.000
```
without_patch

patched

PR #261, opened 2018-01-08
change logging level for failed actions

ohadlevy:
in many cases, I'm seeing trace in the logs where they are not required, for example:
* ERROR -- /executor-dispatcher: ActiveRecord::ConnectionNotEstablished (ActiveRecord::ConnectionNotEstablished)
* invalid worlds found {..}
* /home/foreman/gems/gems/dynflow-0.8.34/lib/dynflow/stateful.rb:35:in `set_state'

while the first two are clear for me, I think the last one is missing printing out the actual exception, but probably can be fixed in another pr.
I'm suggesting to change to WARN so users can change their log level and see the trace, but by default it doesn't log in production the full trace.

### DIFF
--- a/lib/dynflow/delayed_executors/abstract_core.rb
+++ b/lib/dynflow/delayed_executors/abstract_core.rb
@@ -38,7 +38,7 @@ module Dynflow
       def with_error_handling(error_retval = nil, &block)
         block.call
       rescue Exception => e
-        @logger.fatal e.backtrace.join("\n")
+        @logger.warn e.backtrace.join("\n")
         error_retval
       end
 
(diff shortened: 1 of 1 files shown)


iNecas:
I'm ok with this change: also, if you could also put `warn e.message` as part of the PR, it would be the best

ohadlevy:
@iNecas updated - but in that case, I think it makes more sense to change the trace to debug? (updated to reflect that)

adamruzicka:
Fix for the broken tests is here https://github.com/Dynflow/dynflow/pull/262

iNecas:
Thanks @ohadlevy

PR #263, opened 2018-01-09
use current concurrent-ruby(-edge) versions

mmoll:


mmoll:
Seems this doen't work at all. I had hoped the new version of -edge woudl fix the current foreman-tasks test hangs...

PR #264, opened 2018-01-17
Add features needed for reliable recurring

adamruzicka:


lib/dynflow/delayed_plan.rb
```diff
@@ -55,6 +55,11 @@ def to_hash
                         :args_serializer     => @args_serializer.class.name
     end
 
+    def args
```
iNecas: Since we are adding a new public method here, it might be a good idea adding some basic unit test for it as well so that we know others can depend on it.

iNecas:
Thanks @adamruzicka

PR #265, opened 2018-01-25
Add cancelled status for delayed jobs

witlessbird:
This is a workaround rather than a fix. Rather than treating cancellation as an error, a new state needs to be added, but this has too many implications within dynflow as well as foreman-tasks and remote-execution projects.


witlessbird:
Changed the implementation -- introduced 'cancelled' state. Please ignore the initial comment at the top.

adamruzicka:
Could you please add a css class for the cancelled steps so they're easily distinguishable from the "other grey" steps?

lib/dynflow/execution_plan/steps/error.rb
```diff
@@ -1,5 +1,7 @@
 module Dynflow
   module ExecutionPlan::Steps
+    class PlanCancelledError < Dynflow::Error; end
```
adamruzicka: What's the plan with this error class?

witlessbird: None, it's a vestige from the first implementation of "cancelled" state.

witlessbird:
* removed ```PlanCancelledError```
* fixed the failing test

witlessbird:
> Could you please add a css class for the cancelled steps so they're easily distinguishable from the "other grey" steps?

Could you point me to the view you had in mind?

adamruzicka:
A CSS class is assigned to each step depending on its result here[1] and the css definitions are here[2]. It changes the color of the step in execution plan's details view.

[1] - https://github.com/Dynflow/dynflow/blob/master/lib/dynflow/web/console_helpers.rb#L112
[2] - https://github.com/Dynflow/dynflow/blob/master/web/assets/stylesheets/application.css#L72

witlessbird:
* added "warning" css class for cancelled tasks.

It appears that task-status related css classes come from bootstrap css stylesheet, not application.css, unless different styles are used on different pages?

iNecas:
Rebase needed

witlessbird:
* rebased

iNecas:
All green now. @adamruzicka anything else from you side to add?

adamruzicka:
Nope, this looks good

PR #266, opened 2018-01-30
Pin minitest to -> 5.10.0

iNecas:
There is an issue with minitest 5.11.x + Rails 5.1. It should be fixed
in Rails (https://github.com/rails/rails/pull/30800), however, it's not
released yet.

5.11.x gives us issues described in
https://github.com/seattlerb/minitest/issues/730.


adamruzicka:
Thanks @iNecas !

PR #268, opened 2018-02-05
Fix NaN as progress value for WithBulkSubPlans

iNecas:
In case of empty set of sub plans, the aciont with WithBulkSubPlans was
calculating progress as 0./0, which lead to NaN. While some
databases (such as Postgres), were converting this value to 0, on
SQLite, we were getting:

  Sequel::DatabaseError: SQLite3::SQLException: no such column: NaN

The code fixes the WithBulkSubPlans progress calculation + ensuring that
we handle unexpected progress value better.

### DIFF
--- a/lib/dynflow/action/with_bulk_sub_plans.rb
+++ b/lib/dynflow/action/with_bulk_sub_plans.rb
@@ -58,7 +58,7 @@ module Dynflow
 
     # The same logic as in Action::WithSubPlans, but calculated using the expected total count
     def run_progress
-      if counts_set?
+      if counts_set? && total_count > 0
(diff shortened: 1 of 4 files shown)


iNecas:
@dLobatog @adamruzicka this should handle the unexpected `rake aborted` issues in our testing. The thing is that the `Sequel::DatabaseError: SQLite3::SQLException` was raised during the saving of the step. Dynflow's self-healing mechanism assumed, that the SQL exception was caused by issues while connecting to the database and when it failes after several retries, it started terminating the world, that leads to exiting by default. We could disable the exiting when setting up the dynflow world. However, the root cause of the issue is fixed in this PR.

adamruzicka:
Thanks @iNecas !

PR #269, opened 2018-02-06
Reduce the test matrix

iNecas:
Run different databases only on the newest ruby version. Use postgres by
default.


adamruzicka:
How about adding 2.5.0 into the mix?

iNecas:
Makes sense, pushed.

PR #270, opened 2018-02-12
Execution plan life cycle hooks

adamruzicka:
This commit allows actions to hook into the life cycle of an execution plan.
Currently hooks can be triggered when the execution plan goes into:
- `stopped` state (`:stop`)
- `paused` state (`:pause`)
- `stopped` state with `success` result (`:success`)
- `stopped` state with `error`, `warning` or `cancelled` result (`:fail`)

A hook can either override default `execute` method and provide the same
behavior for all events or override separate `on_success`, `on_stop`,
`on_pause` and `on_fail` methods to implement the desired behavior.

Hooks have volatile access to the action, any changes done to the action's
output or input will be lost upon exiting the hook.

Hooks are enabled using the `execution_plan_hooks.use` call and are inherited
when actions are sub-classed. Hooks from the parent action can be disabled using
the `execution_plan_hooks.do_not_use` call in the child action. The order of
execution is not guaranteed.

The hooks are executed in a begin-rescue block, so even if a hook raises an
exception during its run, other hooks should still be run.


iNecas:
Test failures seem related

test/execution_plan_hooks_test.rb
```diff
@@ -0,0 +1,82 @@
...
+        end
+
+        def on_stopped(_execution_plan, _action)
+          Flag.raise!
+          raise "A controlled failure"
+        end
+      end
+
+      class ActionWithHooks < ::Dynflow::Action
+        execution_plan_hooks.use FlagHook, :on => :success
```
iNecas: I wonder if we need a class here vs. just action methods. Seems a bit strange that I need to define both `on_success` in the hook and `:on => :success` here. What about something like this:

```ruby
execution_plan_hooks.run :send_notification, :on => :stopped
```

adamruzicka: And then the shareable parts would be modules which would provide the method and which would register itself when included?

iNecas: Yes, alternatively, we could still stay with the hook classes, but then, I would recommend limiting it to define the `execute` methods, and not using the `on_event`-like method names. It was confusing for me to see the `:on => success` when registering the hook, and seeing also `on_success` in the hook itself.

lib/dynflow/execution_plan/hooks.rb
```diff
@@ -0,0 +1,91 @@
...
+
+        # Runs the registered hooks
+        #
+        # @param execution_plan [ExecutionPlan] the execution plan which triggered the hooks
+        # @param action [Action] the action which triggered the hooks
+        # @param kind [Symbol] the kind of hooks to run, one of {HOOK_KINDS}
+        def run(execution_plan, action, kind)
+          on(kind).each do |hook|
+            begin
+              action.send(hook, kind, execution_plan)
```
iNecas: I wonder, if the `kind` really needs to be passed. If someone needs to distinguish between two hooks, they should probably register different methods for it. What is the case you have in mind that `kind` argument would be useful for?

adamruzicka: No particular use case, dropped that

iNecas:
I still need to test this as a whole, but no other comments code-wise so far.

adamruzicka:
There are still some changes to be done after the schema normalization goes in which will allow us to load only actions which actually have some hooks instead of loading all of them every time.

adamruzicka:
@iNecas this should be ready for another round of reviews

iNecas:
Thanks @adamruzicka

PR #271, opened 2018-02-15
[WIP] Database schema normalization

adamruzicka:
First step toward having normalized database schema. This adds columns for things which were stored in the `data` column.

TODO:
- [ ] make sure this really is backward compatible (i.e. stuff can still be loaded from the `data` column)
- [x] write some backward compatibility tests
- [ ] ~~(optional) consider normalizing `dynflow_coordinator_records`~~
- [ ] ~~(optional) consider normalizing `dynflow_envelopes`~~


lib/dynflow/execution_plan.rb
```diff
@@ -407,7 +407,7 @@ def save
 
     def self.new_from_hash(hash, world)
       check_class_matching hash
-      execution_plan_id = hash[:id]
+      execution_plan_id = hash[:uuid] || hash[:id]
```
iNecas: It's a bit strange, that `to_hash` returns `{:id => "..."}` while `new_from_hash` counts on `{:uuid => "..."}`. What about turning the `{:uuid => "..."}` into `{:id => "..."}` on persistence level so that we can still be consistent here?

iNecas:
Mysql seems to have some issues after this change (see travis).

iNecas:
I wonder if the failure on postgres (https://travis-ci.org/Dynflow/dynflow/jobs/341942121) might be related to any chnange, or this is unrelated… I have not seen this error before (seems like some comparison issue).

iNecas:
In general 👍 Waiting for passing tests + the compatibility layer. I would suggest keeping the envelopes and coordinator records outside of scope of this PR to make it easier to get this changes in.

adamruzicka:
TODO:
- [ ] Blank data column when updating a record using the new schema

adamruzicka:
That was suspiciously easy

adamruzicka:
Rebased to solve migration number conflict

lib/dynflow/persistence_adapters/sequel.rb
```diff
@@ -15,7 +16,7 @@ class Sequel < Abstract
       MAX_RETRIES = 10
       RETRY_DELAY = 1
 
-      attr_reader :db
+      attr_reader :db, :save_as_data
```
iNecas: This is not used anymore

lib/dynflow/persistence_adapters/sequel.rb
```diff
@@ -242,22 +252,37 @@ def migrate_db
...
+        record.reduce({}) do |acc, (key, value)|
+          if SERIALIZABLE_COLUMNS.fetch(table_name, []).include?(key.to_s)
+            acc.merge(key.to_sym => dump_data(value))
+          else
+            acc
+          end
+        end
+      end
+
+      def save(what, condition, value, with_data = true)
```
iNecas: Since `with_data` is used only for testing of the legacy approach, would it be possible to move the preparation of the test data to the test itself, so that the persistence code itself doesn't have to care about saving anything to data from now on?

adamruzicka: It is still used by at least coordinator records. I'll check whether we could do something about it

iNecas:
Tested in multiple scenarios, including foreman_remote_execution. Another thing I was trying was:

```
export DB_CONN_STRING='postgres://postgres@/dynflow_example'
DB=dynflow_example; dropdb -Upostgres $DB; createdb -Upostgres $DB            
bundle exec examples/orchestrate_evented.rb
<ctrl-c>
bundle exec examples/orchestrate_evented.rb
# used dynflow console to cancel the simulated step after resume
```

I also tried foreman_tasks:export_tasks, just in case.

Everything worked as charm. Two pending comments inline, but we are really close to get this in. Looking forward to have that, as that incredibly enhances the readability.

lib/dynflow/persistence_adapters/sequel.rb
iNecas: Oh, in that case that's not that bad. I'm ok with that unless there is some better way.

adamruzicka: I can't think of a solution which would make this significantly better. Let's keep it the way it is for now

iNecas:
There is nothing else I can think of except of thank you @adamruzicka :)

PR #277, opened 2018-03-08
Add multi-queues support

iNecas:
See docs update for more details.

The easiest way to test it out is running

  bundle exec examples/orchestrate.rb

The status page also shows the status of all the queues in the executor
and can be used to show how the system behaves.

### DIFF
--- a/doc/pages/source/documentation/index.md
+++ b/doc/pages/source/documentation/index.md
@@ -950,7 +950,7 @@ To address this case, it's possible to define additional queues tied
 to additional pool of workers dedicated for it. This way, they can be
 processed more independently from the default queue.
 
-To us the queue, one needs to register additional queues when defining
+To use the queue, one needs to register additional queues when defining
 the executor world:
 
 ```ruby
@@ -975,8 +975,9 @@ end
 ```
 
 In the current implementation, it's expected all the executors would
-have the same set of queues defined. In the future implementation,
-dedicated executors with just a subset of queues should be possible.
+have the same set of queues defined. In the future implementation it
+should be possible to have dedicated executors with just a subset of
+queues
 
 ### Middleware
 
--- a/doc/pages/source/documentation/index.md
+++ b/doc/pages/source/documentation/index.md
@@ -940,6 +940,44 @@ Solutions are:
 -   **Offloading computation** - CPU heavy parts can be offloaded to different services
     notifying the suspended actions when the computation is done.
 
+### Multiple queues support
+
+By default, a single queue and a pool of workers is used to process
+all the actions in the system. This can cause some actions to block
+execution of some higher-priority ones in the system.
+
+To address this case, it's possible to define additional queues tied
+to additional pool of workers dedicated for it. This way, they can be
+processed more independently from the default queue.
+
+To us the queue, one needs to register additional queues when defining
+the executor world:
+
+```ruby
+config = Dynflow::Config.new
+config.queues.add(:slow, :pool_size => 5)
+world = Dynflow::World.new(config)
+```
+
+The action to use the queue just needs to override the `queue` method like this:
+
+
+```ruby
+class MyAction < Dynflow::Action
+  def queue
+    :slow
+  end
+
+  def run
+    sleep 60
+  end
+end
+```
+
+In the current implementation, it's expected all the executors would
+have the same set of queues defined. In the future implementation,
+dedicated executors with just a subset of queues should be possible.
+
 ### Middleware
 
 Each action class has chain of middlewares which wrap phases of the action execution.
--- a/lib/dynflow/config.rb
+++ b/lib/dynflow/config.rb
@@ -32,13 +32,49 @@ module Dynflow
         @config.validate(self)
       end
 
+      def queues
+        @queues ||= @config.queues.finalized_config(self)
+      end
+
       def method_missing(name)
         return @cache[name] if @cache.key?(name)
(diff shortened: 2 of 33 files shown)


adamruzicka:
The following migration needs to be added to make this work with normalized schema.

`lib/dynflow/persistence_adapters/sequel_migrations/016_add_queue_column.rb`:
```ruby
Sequel.migration do
  change do
    alter_table(:dynflow_steps) do
      add_column :queue, String
    end
  end
end

```

lib/dynflow/execution_plan/steps/abstract.rb
```diff
@@ -128,11 +134,15 @@ def with_sub_plans?
         false
       end
 
+      def queue
```
adamruzicka: This class already has attribute reader for `queue`. This could be dropped

iNecas:
Sry for missed migration file: I had in on my disk, but forgot to checkout to git, updated now

lib/dynflow/execution_plan/steps/abstract.rb
iNecas: Right, deleted.

lib/dynflow/web/filtering_helpers.rb
```diff
@@ -28,6 +28,12 @@ def filtering_options(show_all = false)
         return @filtering_options
       end
 
+      def load_worlds(executors_only = false)
+        @worlds = world.coordinator.find_worlds(executors_only)
+        @executors = @worlds.select(&:executor?)
+        @clients = @worlds.reject(&:executor?)
```
adamruzicka: Why not split the worlds in one pass with `Enumerable#partition`?

```ruby
@executors, @clients = @worlds.partition(&:executor?)
```

doc/pages/source/documentation/index.md
```diff
@@ -940,6 +940,44 @@ Solutions are:
...
+
+By default, a single queue and a pool of workers is used to process
+all the actions in the system. This can cause some actions to block
+execution of some higher-priority ones in the system.
+
+To address this case, it's possible to define additional queues tied
+to additional pool of workers dedicated for it. This way, they can be
+processed more independently from the default queue.
+
+To us the queue, one needs to register additional queues when defining
```
adamruzicka: `s/us/use`

doc/pages/source/documentation/index.md
```diff
@@ -940,6 +940,44 @@ Solutions are:
...
+
+  def run
+    sleep 60
+  end
+end
+```
+
+In the current implementation, it's expected all the executors would
+have the same set of queues defined. In the future implementation,
+dedicated executors with just a subset of queues should be possible.
```
adamruzicka: Maybe "In the future implementation it should be possible to have dedicated executors with just a subset of queues" would sound more natural?

iNecas: Updated, as a bonus, we got three lines with the same length in a row :)

iNecas:
Related changes leveraging the full power in the remote execution:

- https://github.com/theforeman/foreman_remote_execution/pull/330
- theforeman/foreman-tasks#319
- theforeman/foreman#5320

iNecas:
I've also added support for the active job queues

lib/dynflow/config.rb
```diff
@@ -32,15 +32,51 @@ def validate
...
+      #   By default, it uses global pool_size config option.
+      def add(name, queue_options = {})
+        Utils.validate_keys!(queue_options, :pool_size)
+        name = name.to_sym
+        raise ArgumentError, "Queue #{name} is already defined" if @queues.key?(name)
+        @queues[name] = queue_options
+      end
+
+      def finalized_config(config_for_world)
+        @queues.values.each do |queue_options|
```
adamruzicka: Here we should probably check if the world is an executor world. For client worlds it doesn't make much sense to set queue sizes for them as they can't actually do anything. I know this has no real effect on the functionality, but it might be confusing when shown in the ui.

iNecas: I would not check if the config has executor here, because the executor configuration is using this method, so I would get into endless recursion. However, I updated the code to not pass the metadata about queues if the world doesn't have an executor defined.

lib/dynflow/errors.rb
```diff
@@ -1,5 +1,7 @@
 module Dynflow
   module Errors
+    class ConfigurationError < StandardError; end
```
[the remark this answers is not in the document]
iNecas: Good catch! Updated

lib/dynflow/executors/parallel/core.rb
```diff
@@ -1,16 +1,28 @@
...
-          @pool       = Pool.spawn('pool', reference, pool_size, world.transaction_adapter)
-          @terminated = nil
-          @director   = Director.new(@world)
+        def initialize(world, queues_options)
+          @logger                  = world.logger
+          @world                   = Type! world, World
+          @queues_options          = queues_options
+          @pools                   = {}
+          @terminated              = nil
+          @director                = Director.new(@world)
```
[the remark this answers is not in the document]
iNecas: Purely esthetic reasons :paintbrush: :art:… :)

I guess there were some additional long variables present in the process, that I've removed later, but haven't updated the indentation. Fixed now.

iNecas:
All comments should be addressed now.

adamruzicka:
Big thanks @iNecas !

PR #278, opened 2018-03-09
Fixes #22837 - postpone validation checks in Rails environment

iNecas:
Otherwise, the actions performed during the validity check could lead to
issues like this:

    RuntimeError: The Dynflow world was not initialized yet. If your plugin
    uses it, make sure to call Rails.application.dynflow.require! in some
    initializer


adamruzicka:
Test failures related, `DummyWorld` used in testing doesn't implement `perform_validity_checks`

adamruzicka:
@iNecas, @ehelms thanks!

PR #279, opened 2018-03-09
Fixes #22839 - defensive cleanup when running user-defined code

iNecas:
In case we run user code inside Dynflow threads, we should make sure to
clean common mess, such as leaking ActiveRecord connections.

### DIFF
--- a/lib/dynflow/execution_plan/hooks.rb
+++ b/lib/dynflow/execution_plan/hooks.rb
@@ -60,12 +60,16 @@ module Dynflow
         # @param action [Action] the action which triggered the hooks
         # @param kind [Symbol] the kind of hooks to run, one of {HOOK_KINDS}
         def run(execution_plan, action, kind)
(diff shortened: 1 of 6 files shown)


lib/dynflow/execution_plan/hooks.rb
```diff
@@ -62,7 +62,9 @@ def dup
         def run(execution_plan, action, kind)
           on(kind).each do |hook|
             begin
-              action.send(hook, execution_plan)
+              Executors.run_user_code do
```
adamruzicka: I know it is better to be on the safe side, but is it really necessary to do the cleanup after running every single hook? I mean there could be more hooks bound to a single execution plan state change

iNecas: Updated

adamruzicka:
@iNecas, @ehelms thanks!

PR #280, opened 2018-04-11
Fixes #23156 - handle unexpected plan states

iNecas:
When a process is terminated while the delayed executor is processing,
it can lead to the situation, where the plan is in weird state (such as
:planning, or :stopped), while the delayed record is still present.

This patch fixes this by returning the `:planning` state back to
`:scheduled`, and in case of other states, just skipping that, because
in that case, other resiliency mechanisms would already handled that, so
we need just to make sure we delete the delayed record at the end.

### DIFF
--- a/lib/dynflow/delayed_executors/abstract_core.rb
+++ b/lib/dynflow/delayed_executors/abstract_core.rb
@@ -46,8 +46,12 @@ module Dynflow
       def process(delayed_plans, check_time)
         processed_plan_uuids = []
         delayed_plans.each do |plan|
+          fix_plan_state(plan)
           with_error_handling do
-            if !plan.start_before.nil? && plan.start_before < check_time
+            if plan.execution_plan.state != :scheduled
+              # in case the previous process was terminated after running the plan, but before deleting the delayed plan record.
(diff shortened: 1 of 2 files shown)


lib/dynflow/delayed_executors/abstract_core.rb
```diff
@@ -46,8 +46,12 @@ def with_error_handling(error_retval = nil, &block)
       def process(delayed_plans, check_time)
         processed_plan_uuids = []
         delayed_plans.each do |plan|
+          fix_plan_state(plan)
```
adamruzicka: Shouldn't we do this conditionally instead of always calling the method and doing the check there?

iNecas: The reason I did it this way was so that the condition, as well as the fix is on the same place: otherwise, the logic would be split into two parts, which is something I would like to avoid

adamruzicka: Ok, fair enought

adamruzicka:
Thanks @iNecas !

PR #281, opened 2018-04-25
Clear AR connections only if there are no open transactions

adamruzicka:
Always clearing the connections broke tests in REX

### DIFF
--- a/lib/dynflow/executors.rb
+++ b/lib/dynflow/executors.rb
@@ -8,9 +8,10 @@ module Dynflow
     # we should wrap it with this method, and we can ensure here to do
     # necessary cleanup, such as cleaning ActiveRecord connections
     def self.run_user_code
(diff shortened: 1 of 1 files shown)


lib/dynflow/executors.rb
```diff
@@ -8,9 +8,10 @@ module Executors
     # we should wrap it with this method, and we can ensure here to do
     # necessary cleanup, such as cleaning ActiveRecord connections
     def self.run_user_code
+      clear_connections = defined?(::ActiveRecord) && ActiveRecord::Base.connection.open_transactions.zero?
```
iNecas: Tests are failing, probably because we actually don't set the connection. Adding `ActiveRecord::Base.connected?` should help.

adamruzicka:
We're green now

iNecas:
Thanks @adamruzicka

PR #283, opened 2018-05-04
Fixes #23501 - ensure to initialize dynflow in executor

iNecas:
In development environment, the dynflow lazy loading could lead to not
initializing the executor.


iNecas:
Tests should be fixed now

adamruzicka:
Thanks @iNecas !

PR #284, opened 2018-05-14
Refs #21358 - run on_failure hook on cancelled and warning states

iNecas:
The cancelled state on steps introduced in a023b2907 caused the
regression, where `on_failure` hooks were not run on `cancelling`.

I've also noticed we don't run them on `warning` as well. After the
change, we do run it in both cases.


adamruzicka:
Thanks @iNecas !

PR #286, opened 2018-06-08
Allow middlewares to be executed around hooks

adamruzicka:
Rationale:
When we moved repetition triggering for recurring actions into a hook, suddenly it wasn't being executed as the original user and in the original taxonomies. This will allow us to fix it once by allowing the middlewares to restore user and taxonomies before running the hook instead of having to deal with the same issue for every recurring action separately.


iNecas:
Nice and clean. Tested and works well. Thanks @adamruzicka, new dynflow release to follow.

PR #287, opened 2018-06-11
Set job.provider_job_id

iNecas:
This makes it easier to get to the execution plan from the job object.

During testing, I've hit a corner case, where plan.root_plan_step was nil: it's
better to be safe then sorry here.


adamruzicka:
Thanks @iNecas !

PR #288, opened 2018-06-22
Collect some basic telemetry

adamruzicka:
## List of collected metrics
gauge dynflow_active_workers:
  Represents the number of currently active workers.
  tags: queue, world id

counter dynflow_worker_events:
  Represents the number of events processed by a worker.
  tags: queue, world id, worker name

gauge dynflow_active_execution_plans:
  Represents the number of currently active (state != stopped)
  execution plans.
  tags: label, state, world id

counter dynflow_finished_execution_plans:
  Represents the number of finished execution plans (state = stopped).
  tags: label, result, world id

histogram dynflow_step_real_time:
  Collects statistics about steps' real time attribute.
  tags: action, phase

histogram dynflow_step_execution_time:
  The same as dynflow_step_real_time histogram but for execution time.

counter dynflow_connector_envelopes:
  The number of envelopes send and received by a connector
  tags: direction, world id

The companion code for integration with Foreman lives here https://github.com/adamruzicka/foreman/tree/dynflow-telemetry


lib/dynflow/action.rb
```diff
@@ -584,6 +590,18 @@ def self.singleton_locked?(world)
...
       end
     end
+
+    def update_step_telemetry(phase)
+      Dynflow::Telemetry.with_instance do |t|
+        t.observe_histogram(:dynflow_step_real_time,
+                            @step.real_time,
+                            :action => self.class.to_s, :phase => phase)
+        t.observe_histogram(:dynflow_step_execution_time,
+                            @step.execution_time,
```
adamruzicka: When I looked at the collected data in Prometheus, the histograms looked weird, but maybe I was just interpreting it wrong. I'll need to wrap my head around the whole buckets thing so this may need tweaking.

iNecas:
I think it would make sense to bring the statsd on in.

lib/dynflow/director/execution_plan_manager.rb
```diff
@@ -99,6 +99,17 @@ def start_finalize
       def finish
         execution_plan.execution_history.add('finish execution', @world.id)
         @execution_plan.update_state(execution_plan.error? ? :paused : :stopped)
+
+        Dynflow::Telemetry.with_instance do |t|
+          @execution_plan.finalize_steps.each do |step|
+            t.observe_histogram(:dynflow_step_real_time,
+                                step.real_time,
+                                :action => self.class.to_s, :phase => 'finalize')
```
iNecas: shouldn't it be `step.action_class` or somethign like that?

adamruzicka:
Currently the active_execution_plans metric makes sense only in single-world deployment. In case there are more worlds (for example a client producing work and executor consuming it), count of active plans for one world grows and decreases for another.

image

For this to work we will probably have to stop scoping active execution plans per-world.

lzap:
Nice, let me do quick look through the codebase.

What is "world id" exactly and why do you use it as a parameter?

examples/example_helper.rb
```diff
@@ -27,6 +28,14 @@ def persistence_conn_string
...
+    def telemetry_adapter
+      if (host = ENV['TELEMETRY_STATSD_HOST'])
+        Dynflow::TelemetryAdapters::StatsD.new host
+      else
+        Dynflow::TelemetryAdapters::Dummy.new
+      end
+    end
+
     def persistence_adapter
       Dynflow::PersistenceAdapters::Sequel.new persistence_conn_string
```
lzap: Persistence layer can be easily a bottleneck too, if you can try to inject some measurements into sequel (SQL query duration for example).

lib/dynflow/action.rb
lzap: There was a bug fixed in core recently, units were incorrect: https://github.com/theforeman/foreman/pull/5874/files

Histogram can't be easily plotted in time, it's better to look on them aggregated https://www.google.cz/search?q=histogram

But if you need to look thorugh the time, it's usually better just to pick several buckets and draw them together as line graph. Prometheus can calculate approximate percentiles which is usually very good way of interpreting data (99th, 95th percentile): https://kausal.co/blog/prometheus-histograms-multiple-queries/

lib/dynflow/execution_plan.rb
```diff
@@ -547,6 +554,21 @@ def unlock_all_singleton_locks!
...
+    def toggle_telemetry_state(original, new)
+      return if original == new
+      @label = root_plan_step.action_class if @label.nil?
+      Dynflow::Telemetry.with_instance do |t|
+        t.set_gauge(:dynflow_active_execution_plans, '-1',
+                    telemetry_common_options.merge(:state => original)) unless original.nil?
+        t.set_gauge(:dynflow_active_execution_plans, '+1',
+                    telemetry_common_options.merge(:state => new)) unless new.nil?
+      end
+    end
```
lzap: Do you ever send absolute number? I mean, relative is the correct approach but is it by default set to zero? I hope so

lib/dynflow/executors/parallel/pool.rb
```diff
@@ -46,12 +46,17 @@ def delete(execution_plan_id)
...
-        def initialize(core, name, pool_size, transaction_adapter)
+        def initialize(world, core, name, pool_size, transaction_adapter)
+          @world = world
           @name = name
           @executor_core = core
           @pool_size     = pool_size
-          @free_workers  = Array.new(pool_size) { |i| Worker.spawn("worker-#{i}", reference, transaction_adapter) }
           @jobs          = JobStorage.new
+          @free_workers  = Array.new(pool_size) do |i|
+            name = "worker-#{i}"
```
lzap: This name goes to telemetry, it is worth advertising it via UNIX process name so it also appears in `ps ax` output. In case there is some worker struggling this would allow you to identify the process and do something (kill it, finding RSS memory etc). Ideal naming convention would be: `dynlfow [worker-1]` or similar.

lib/dynflow/telemetry_adapters/abstract.rb
```diff
@@ -0,0 +1,71 @@
+module Dynflow
+  module TelemetryAdapters
+    class Abstract
+      # Default buckets to use when defining a histogram
+      DEFAULT_BUCKETS = [10, 50, 200, 1000, 15_000].freeze
```
lzap: These need to be in seconds.

examples/example_helper.rb
adamruzicka: Good point, I'll try looking into it

lib/dynflow/execution_plan.rb
adamruzicka: No, just increment and decrement by 1. From what I've seen when implementing this the gauge starts at 0

lib/dynflow/executors/parallel/pool.rb
adamruzicka: I'm afraid the workers run as fibers inside the executor process, so all the workers show up just as `dynflow_executor` in `ps ax`

adamruzicka:
> What is "world id" exactly and why do you use it as a parameter?

A world is one Dynflow runtime. In Rails each thread which triggers actions in Dynflow is a world and each executor is a world.

I used the id as a parameter mainly to be able to tell "how much work is executor X doing", instead of just collecting metrics about "the whole Dynflow together is doing this much work".

lib/dynflow/executors/parallel/pool.rb
lzap: Oh ok I thought this is a multi-process environment.

lib/dynflow/execution_plan.rb
lzap: Great!

lib/dynflow/telemetry_adapters/abstract.rb
iNecas: This were the buckets so far I was testing with, that seemed like good distrubution for the tasks timing:

```
[0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 20, 30, 60, 120, 300, 600, 1200]
```

lib/dynflow/director/execution_plan_manager.rb
```diff
@@ -99,6 +99,17 @@ def start_finalize
       def finish
         execution_plan.execution_history.add('finish execution', @world.id)
         @execution_plan.update_state(execution_plan.error? ? :paused : :stopped)
+
+        Dynflow::Telemetry.with_instance do |t|
+          @execution_plan.finalize_steps.each do |step|
+            t.observe_histogram(:dynflow_step_real_time,
+                                step.real_time,
```
iNecas: Needs to be in ms, so `*1000`

lib/dynflow/director/execution_plan_manager.rb
```diff
@@ -99,6 +99,17 @@ def start_finalize
...
+                                :action => step.action_class.to_s, :phase => 'finalize')
+            t.observe_histogram(:dynflow_step_execution_time,
+                                step.execution_time,
```
iNecas: Here *1000 as well

lib/dynflow/action.rb
```diff
@@ -584,6 +590,18 @@ def self.singleton_locked?(world)
         false
       end
     end
+
+    def update_step_telemetry(phase)
+      Dynflow::Telemetry.with_instance do |t|
+        t.observe_histogram(:dynflow_step_real_time,
+                            @step.real_time,
```
iNecas: As also @lzap mentioned, this needs to be `*1000`

lib/dynflow/executors/parallel/pool.rb
```diff
@@ -46,12 +46,17 @@ def delete(execution_plan_id)
...
+            telemetry_options = { :world => @world.id, :worker => name, :queue => @name }
```
iNecas: This should use `telemetry_options` method as a base, shouldn't it?

adamruzicka: Good catch

adamruzicka:
Addressed comments with seconds vs miliseconds and rebased, still need to figure the right spots to measure interactions with the db

lib/dynflow/telemetry_adapters/abstract.rb
lzap: Yeah all credit goes to @iNecas who found it. I had no idea.

FYI statsd does not have any notion of buckets, this is for prometheus exporter, you don't need them in your codebase, in core it is part of stack because then there is a YAML exporter generator, here you have your YAML hardcoded - so drop this line, it's useless.

adamruzicka: The yaml from the repository is used only when running "standalone Dynflow". When Dynflow is used with Foreman, these values should still get picked up by the yaml exporter generator. That's how the hardcoded yaml in this repo was created in the first place

lib/dynflow/director/execution_plan_manager.rb
lzap: Yep statsd is always in ms.

lzap:
I suggest you to try this with pcp-mmvstatsd as documented in: https://theforeman.org/2018/07/monitoring-and-telemetry-of-foreman-118.html

This will be likely preferred integration downstream, it won't give you buckets tho. I can add percentiles tho: p99, p95, p90.

statsd_mapping.conf
```diff
@@ -0,0 +1,61 @@
+---
```
iNecas: What about introducing `extras` dir for files like this?

lib/dynflow/telemetry.rb
```diff
@@ -0,0 +1,55 @@
...
+
+          # Execution plan related
+          t.add_gauge   :dynflow_active_execution_plans, 'The number of active execution plans',
+                        [:action, :world, :state]
+          t.add_counter :dynflow_finished_execution_plans, 'The number of execution plans',
+                        [:action, :world, :result]
+
+          # Step related
+          # TODO: Configure buckets in a sane manner
+          t.add_histogram :dynflow_step_real_time, 'The real time spent executing a step',
```
iNecas: The description for both `real` and `execution` time is the same: there should be probably the difference explained.

lib/dynflow/action.rb
```diff
@@ -545,10 +548,13 @@ def execute_run(event)
           check_serializable :output
         end
 
+        # Only update the telemetry data if the step stopped for good
+        update_step_telemetry 'run' if [:success, :skipped].include? self.state
```
iNecas: This is called too early, as the metadata is recalculated here https://github.com/Dynflow/dynflow/blob/4c3e72b187037bb9e1607906518bb445ea774717/lib/dynflow/execution_plan/steps/abstract_flow_step.rb#L16, so it actually has values from previous run. 

Also, the execution time should be IMO updated every time the run is executed, but just with the time it took to execute it. The right place probabely should be in the `with_metadata_calculation` method.

examples/example_helper.rb
adamruzicka: Which level of detail would make sense? Several options come to mind

- Query duration with details - "There was a SELECT/INSERT/DELETE which took XYms"
- Query duration - "There was an unspecified SQL query which took XYms"
- Duration per execution plan - "When executing an execution plan, Dynflow spent XYms in the database", closest to `:http_request_db_duration` in Foreman

We could probably get even more details but I'm not sure if it would be useful in any way.

lib/dynflow/debug/persistence.rb
```diff
@@ -0,0 +1,40 @@
+module Dynflow
```
iNecas: I would probably call it `telemetry/persistence.rb`

iNecas:
One last nitpick + travis is failing due to rubocop, but perhaps even some actual test failures.

adamruzicka:
Addressed the comment and fixed rubocop+tests and everything is green now

iNecas:
Tested and work well. Thanks @adamruzicka and @lzap. Happy telemetrying :)

PR #289, opened 2018-06-22
Add frozen state to delayed plans and do not execute them

jlsherrill:


lib/dynflow/persistence_adapters/sequel.rb
```diff
@@ -124,7 +124,7 @@ def find_old_execution_plans(age)
       def find_past_delayed_plans(time)
         table_name = :delayed
         table(table_name)
-          .where(::Sequel.lit('start_at <= ? OR (start_before IS NOT NULL AND start_before <= ?)', time, time))
+          .where(::Sequel.lit('start_at <= ? OR (start_before IS NOT NULL AND start_before <= ?) and frozen != \'t\'', time, time))
```
jlsherrill: I wasn't sure exactly if this was needed, but it seemed like it.

jlsherrill:
Still have tests to fix and add, but this general approach seems to work.  I can freeze/or unfreeze a delayed job with:

To freeze:
```
 r = ForemanTasks::RecurringLogic.find(10);
ForemanTasks.dynflow.world.persistence.freeze_delayed_plan(r.tasks.where(:state => :scheduled).first.execution_plan.id, true)
```

To unfreeze:
```
ForemanTasks.dynflow.world.persistence.freeze_delayed_plan(r.tasks.where(:state => :scheduled).first.execution_plan.id, false)
```

and it starts/stops as expected.  One downside/issue i see is that it seems to run all the missed scheduled runs.  This may be intentional or not, but i think we need some way to change the behavior (I think it should NOT run those by default)..

lib/dynflow/persistence.rb
```diff
@@ -93,6 +93,12 @@ def save_delayed_plan(delayed_plan)
       adapter.save_delayed_plan(delayed_plan.execution_plan_uuid, delayed_plan.to_hash)
     end
 
+    def freeze_delayed_plan(execution_plan_id, frozen = true)
+      plan = load_delayed_plan(execution_plan_id)
+      plan.frozen = frozen
+      save_delayed_plan(plan)
+    end
```
jlsherrill: I can split this into a freeze and unfreeze method if preferred.

lib/dynflow/delayed_executors/abstract_core.rb
```diff
@@ -46,6 +46,7 @@ def with_error_handling(error_retval = nil, &block)
       def process(delayed_plans, check_time)
         processed_plan_uuids = []
         delayed_plans.each do |plan|
+          next if plan.frozen
```
adamruzicka: Is this really needed? If I'm reading the [delayed plan checking code](https://github.com/Dynflow/dynflow/pull/289/files#diff-59489d3bacadc3676de871dc37b984c7R127)  correctly, only plans which aren't frozen  should get loaded from the db and passed here

lib/dynflow/persistence_adapters/sequel.rb
adamruzicka: Do all the DBs we support use the same value (`'t'`) to store truthy values? Maybe turning this into something like the following would be safer

```ruby
table(table_name)
  .where(:frozen => false)
  .where(::Sequel.lit('start_at <= ? OR (start_before IS NOT NULL AND start_before <= ?)', time, time))
```

lib/dynflow/delayed_executors/abstract_core.rb
jlsherrill: That change alone didn't seem to handle it, but I will retest, its possible i didn't stop spring or something

lib/dynflow/persistence.rb
iNecas: either split or `set_freeze`, as it would indicate more that it's about setting the `frozen` flag, not just setting it to true (which it reads like right now)

adamruzicka:
>  One downside/issue i see is that it seems to run all the missed scheduled runs. This may be intentional or not, but i think we need some way to change the behavior (I think it should NOT run those by default).

Couly you elaborate on this?

jlsherrill:
> Could you elaborate on this?

Actually i rested this, what is happening is that after unfreezing a delayed_plan, the task gets executed in a strange pattern.  Take this instance:
1) Created a delayed plan using: "*/2 * * * *"  (every two minutes)
2)  Let it execute a few times, last execution time is: 2018-06-25 12:50:03 -0400
3) I wait a few minutes and unfreeze it, I then see these tasks get created with started_at times of:

2018-06-25 12:54:48 -0400
2018-06-25 12:55:04 -0400
2018-06-25 12:56:04 -0400
2018-06-25 12:58:04 -0400
2018-06-25 13:00:04 -0400

The first one here seems like its running the 'past due' execution, but i can't explain the 2nd entry.  You see the 3rd and 4th re-establish the 'every two minute' pattern

lib/dynflow/delayed_executors/abstract_core.rb
jlsherrill: Yes, this appears to be required still.

lib/dynflow/persistence_adapters/sequel.rb
jlsherrill: I assume you meant something like:
```
          .where(::Sequel.lit('start_at <= ? OR (start_before IS NOT NULL AND start_before <= ?) and frozen != ?', time, time, true))
```

jlsherrill:
Another thing i noticed was that an every minute job (* * * * *) seemed to execute 2x a minute, something seemed not quite right.

lib/dynflow/persistence_adapters/sequel.rb
jlsherrill: added

adamruzicka: Chained where conditions should get joined by ands into one, but I'm ok with your solution

jlsherrill: i see, i misread your comment.  Thanks!

iNecas:
@jlsherrill since the recurring logic is actually handled in foreman-tasks, you should be able to update the future time after unfreezing there: maybe we should let updating the `start_at` time in the `set_delayed_plan_frozen` method

adamruzicka:
@jlsherrill Are you expecting any more changes to this?

jlsherrill:
@adamruzicka i don't believe so, but we can wait a bit till https://github.com/theforeman/foreman-tasks/pull/350 is Good to go if you want

iNecas:
No other comments from me.

adamruzicka:
Thanks @jlsherrill !

PR #291, opened 2018-07-02
Fixes #24119 - ensure delayed executor is preserved after invalidation

iNecas:
We do two things to achieve this:

  1. re-initialize the delayed executor if some worlds were invalidated

  2. don't use ping cache at executors startup - this way we handle the
  case, when the executor gets force killed and the timeout doesn't
  expire before the new executor goes up. I've increased the ping
  timeout to avoid false negatives

Steps to reproduce:

1. run executor with rails environment (with foreman-tasks)
2. force kill the executor
3. run executor again

Without this patch: the old executor doesn't get invalidated at startup and even it it gets (when waiting for a minute between step 2 and 3, it doesn't get the delayed executor started.

### DIFF
--- a/lib/dynflow/dispatcher/client_dispatcher.rb
+++ b/lib/dynflow/dispatcher/client_dispatcher.rb
@@ -138,7 +138,7 @@ module Dynflow
                             (on ~Event do |event|
                                find_executor(event.execution_plan_id)
                              end),
-                            (on Ping.(~any) | Status.(~any, ~any) do |receiver_id, _|
+                            (on Ping.(~any, ~any) | Status.(~any, ~any) do |receiver_id, _|
                                receiver_id
                              end)
         envelope = Envelope[request_id, client_world_id, executor_id, request]
@@ -243,6 +243,7 @@ module Dynflow
       # @return [Concurrent::Future] the future tracking the request
       def with_ping_request_caching(request, future)
         return yield unless request.is_a?(Dynflow::Dispatcher::Ping)
+        return yield unless request.use_cache
 
         if @ping_cache.fresh_record?(request.receiver_id)
           future.success(true)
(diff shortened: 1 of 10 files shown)


lib/dynflow/world.rb
```diff
@@ -63,14 +55,38 @@ def initialize(config)
...
       end
+      post_initialization
+    end
+
+    # performs steps once the executor is ready and invalidation of previous worls is finished.
+    # Needs to be indempotent, as it can be called several times (expecially when auto_validity_check
+    # if false, as it should be called after `perform_validity_checks` method)
+    def post_initialization
+      @delayed_executor         ||= try_spawn(:delayed_executor, Coordinator::DelayedExecutorLock)
+      @execution_plan_cleaner   ||= try_spawn(:execution_plan_cleaner, Coordinator::ExecutionPlanCleanerLock)
```
adamruzicka: I would expect rubocop to complain about the spacing here

lib/dynflow/dispatcher/client_dispatcher.rb
```diff
@@ -243,6 +243,7 @@ def resolve_tracked_request(id, error = nil)
       # @return [Concurrent::Future] the future tracking the request
       def with_ping_request_caching(request, future)
         return yield unless request.is_a?(Dynflow::Dispatcher::Ping)
+        return yield if !request.use_cache
```
adamruzicka: I'd prefer `unless` over `if !`

iNecas: Me too :)

iNecas:
Updated

iNecas:
Tests are green

iNecas:
Fixed this issue by registering just after the executor was started: I've also added there termination in case the record is missing, which could only happen if we false negatively terminated some world, that took too long to respond.

iNecas:
Feel free to merge it then :)

adamruzicka:
Thanks @iNecas !

PR #292, opened 2018-07-17
Handle situation when execution plan gets deleted

iNecas:
Before this patch, we were considering every persistence issue as
something that we could retry on and the only resolution if that was not
successful, was executor termination.

With this patch, we keep the retry mechanism just for connection issues
and for other persistence issues, we just clean up the runtime.


adamruzicka:
Thanks @iNecas !

PR #293, opened 2018-07-27
Fixes #24435 - Remove round robin from the job storage

iNecas:
This was leading to some unexpected behavior: using simple FIFO instead.

This PR aslo adds simple benchmarking script to compare the behavior: ideal in
combination with the telemetry.

### DIFF
--- a/lib/dynflow/executors/parallel/pool.rb
+++ b/lib/dynflow/executors/parallel/pool.rb
@@ -4,19 +4,15 @@ module Dynflow
       class Pool < Actor
         class JobStorage
           def initialize
-            @round_robin = RoundRobin.new
-            @jobs        = Hash.new { |h, k| h[k] = [] }
+            @jobs = []
           end
 
           def add(work)
-            @round_robin.add work.execution_plan_id unless tracked?(work)
-            @jobs[work.execution_plan_id] << work
+            @jobs << work
           end
 
           def pop
-            return nil if empty?
-            execution_plan_id = @round_robin.next
-            @jobs[execution_plan_id].shift.tap { delete execution_plan_id if @jobs[execution_plan_id].empty? }
+            @jobs.shift
           end
 
           def queue_size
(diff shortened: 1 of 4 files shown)


iNecas:
Some comparison charts, showing steps real time buckets le 10, 20, 30 and Inf during the execution of the benchmark script. You can notice that in the variant without round robin, much more steps finish with shorter real time. The overall time of the benchmark is just slightly faster (86s vs 88s), but the main difference is in the distribution of duration values.

With round robin https://snapshot.raintank.io/dashboard/snapshot/udI37iI6TL07nlMxhM4RPVlOuNhrk2pr?orgId=2

Without round robin https://snapshot.raintank.io/dashboard/snapshot/0p5BNwHxzhch0DgfxlH9R4Nrs6lE0R0M?orgId=2

examples/benchmark.rb
```diff
@@ -0,0 +1,217 @@
...
+                   ping_interval: 0.5,
+                   max_iterations: 2 }
+
+require 'optparse'
+require 'pp'
+
+OptionParser.new do |opts|
+  opts.banner = "Usage: benchmark.rb [options]"
+
+  opts.on("c", "--connection-string [CONNECTION_STRING]", "Database connection string (default #{options[:connection_string]}") do |v|
```
adamruzicka: `s/c/-c/`

examples/example_helper.rb
```diff
@@ -63,4 +63,75 @@ def terminate
...
+    def run_server
+      world = ExampleHelper.create_world do |config|
+        config.persistence_adapter = persistence_adapter
+        config.connector           = connector
+      end
+      run(world)
+    end
+
+    def client
+      @client ||= ExampleHelper.create_world do |config|
```
adamruzicka: This "create world, assign persistence adapter and connector" pattern repeats a lot, also there's another `#client` a bit below https://github.com/Dynflow/dynflow/pull/293/files#diff-1d1e81c6ffa6ce36582fbb4c1e6c2e59R126

examples/benchmark.rb
```diff
@@ -0,0 +1,217 @@
...
+require 'tmpdir'
+require 'stringio'
+
+class SampleAction < Dynflow::Action
+  def plan(sub_actions_count: 2,
+           step_duration: 0.5,
+           step_duration_range: nil,
+           ping_interval: 0.5,
+           max_iterations: 2)
+    id = rand(1e10)
```
adamruzicka: Can't we just take a part of the execution plan's uuid?

lib/dynflow/executors/parallel/pool.rb
```diff
@@ -4,19 +4,16 @@ class Parallel < Abstract
...
-            @round_robin.add work.execution_plan_id unless tracked?(work)
-            @jobs[work.execution_plan_id] << work
+            @jobs << work
           end
 
           def pop
             return nil if empty?
-            execution_plan_id = @round_robin.next
-            @jobs[execution_plan_id].shift.tap { delete execution_plan_id if @jobs[execution_plan_id].empty? }
+            @jobs.shift
```
adamruzicka: The `return nil if empty?` is functionally equivalent to `[].shift`. Some quick testing also shows that just shifting is also a tiny bit faster in this case.

iNecas: Updated

iNecas:
Everything should be ready here

adamruzicka:
Thanks @iNecas !

PR #295, opened 2018-09-06
Fixes #24829 - Fixes setting pool size when using rails

adamruzicka:
Before this commit, the setting would be simply overridden by 5.


iNecas:

PR #296, opened 2018-09-21
Test update concurrent-ruby to 1.1.3

adamruzicka:
Examples seem to work, certain tests get stuck.

### DIFF
--- a/dynflow.gemspec
+++ b/dynflow.gemspec
@@ -21,8 +21,8 @@ Gem::Specification.new do |s|
   s.add_dependency "multi_json"
   s.add_dependency "apipie-params"
   s.add_dependency "algebrick", '~> 0.7.0'
-  s.add_dependency "concurrent-ruby", '= 1.1.0.pre2'
-  s.add_dependency "concurrent-ruby-edge", '= 0.4.0.pre2'
+  s.add_dependency "concurrent-ruby", '~> 1.1.3'
+  s.add_dependency "concurrent-ruby-edge", '~> 0.4.1'
   s.add_dependency "sequel", '>= 4.0.0'
 
   s.add_development_dependency "rake"
--- a/dynflow.gemspec
+++ b/dynflow.gemspec
@@ -21,8 +21,8 @@ Gem::Specification.new do |s|
   s.add_dependency "multi_json"
   s.add_dependency "apipie-params"
   s.add_dependency "algebrick", '~> 0.7.0'
-  s.add_dependency "concurrent-ruby", '~> 1.0.0'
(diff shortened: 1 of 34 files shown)


iNecas:
Nice @adamruzicka, let's continue with the tests, but looks promising…

adamruzicka:
This leaves us with two failing tests:
- [ ] `abnormal_states_recovery_test` which currently deadlocks
- [ ] `executor_test` which just hangs

adamruzicka:
Tests are passing for me locally, let's see what Mr. T say. :crossed_fingers:

adamruzicka:
And we're :green_heart:

iNecas:
Very nice. I don't want to get this in right now due to upcoming 1.20 release, but once that's over, we can get this in.

FYI @pitr-ch - the new c-r relesae seems good so far… :)

iNecas:
And I think @mmoll will be also happy about this PR :)

adamruzicka:
That makes sense, I'd like to give it a spin in a production environment before this goes in.

iNecas:
That's a good point: it's also wroth checking if we're not using any of the old API in foreman plugins and if so, ideally having some limited compatibility layer here to allow a smooth transition, if possible

dynflow.gemspec
```diff
@@ -21,8 +21,8 @@ Gem::Specification.new do |s|
   s.add_dependency "multi_json"
   s.add_dependency "apipie-params"
   s.add_dependency "algebrick", '~> 0.7.0'
-  s.add_dependency "concurrent-ruby", '~> 1.0'
-  s.add_dependency "concurrent-ruby-edge", '~> 0.2.0'
+  s.add_dependency "concurrent-ruby", '= 1.1.0.pre2'
+  s.add_dependency "concurrent-ruby-edge", '= 0.4.0.pre2'
```
mmoll: is this still needed?

iNecas: I guess it's needed until the c-r 1.1 is in pre-release

adamruzicka: We still use Actors which live in -edge

mmoll: I see :+1:

pitr-ch:
Great! :) Thanks for trying it out.

mmoll:
What's the status here? :)

iNecas:
After conflicts are resolved and tests are green, I'm ok to hit the merge button

adamruzicka:
And it's :green_heart:

iNecas:
Everything I'v tried seems to work just fine: I don't see any other reason to hold this back any longer. Thanks @adamruzicka

iNecas:
And of course @pitr-ch for great work on concurrent-ruby gem

PR #297, opened 2018-09-24
Fixes #25021 - add more termination timeouts

iNecas:
This patch adds a timeout for some connector events, waiting for
termination hooks, as well as a global termination timeout to make sure we
don't get stuck forever.

I've added the extraction of the method to separate commit, so that it's easier
to see the changes in the orinial method.


iNecas:
Anything to add before merge?

adamruzicka:
No, just forgot to push the button

adamruzicka:
Thanks @iNecas !

PR #299, opened 2018-10-23
Consider tasks which fail planning as planned in bulk action

adamruzicka:


iNecas:
Tested and worked well. Thanks @adamruzicka

PR #300, opened 2018-10-23
Add method missing to logger wrapper

iNecas:
Otherwise, the wrappers are not transparent enough, and the users can get issues
with some methods missing.

### DIFF
--- a/lib/dynflow/logger_adapters/simple.rb
+++ b/lib/dynflow/logger_adapters/simple.rb
@@ -46,6 +46,18 @@ module Dynflow
           end
         end
 
+        def respond_to_missing?(method_name, include_private = false)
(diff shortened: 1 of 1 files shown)


lib/dynflow/logger_adapters/simple.rb
```diff
@@ -46,6 +46,18 @@ def initialize(logger, prog_name)
           end
         end
 
+        def respond_to_missing?(method_name, include_private = false)
+          @logger.respond_to?(metohd_name, include_private) || super
```
adamruzicka: `s/ohd/hod`

adamruzicka:
Thanks @iNecas, @lzap !

PR #301, opened 2018-11-09
Make sure we always terminate

adamruzicka:


adamruzicka:
This should be now ready for review, only https://github.com/Dynflow/dynflow/pull/301/commits/a41ac592b0d4fa3caa1c82182a5ef76b5b2c2e9c is really essential

iNecas:
Tested and works well. Thanks @adamruzicka: btw. do we have a BZ and redmine issue to track this problem, so that it's easier to make sure the right versions get the patches?

adamruzicka:
We do have a BZ[1] now.

[1] - https://bugzilla.redhat.com/show_bug.cgi?id=1654975

iNecas:
Note for myself: redmine issue here https://projects.theforeman.org/issues/25593

PR #302, opened 2018-11-13
Use correct logger in Rails daemon

adamruzicka:
`#dynflow_logger` is not defined on `world`


iNecas:
I was testing this and I can confirm it fixes the issue of `undefined method `dynflow_logger'`. There is still some issue of the logs not actually showing up for some reason (it seems like the file is not flushing to disk :/, but that seems orthogonal to this problem. Thanks @adamruzicka

iNecas:
Note for myself: link to bz https://bugzilla.redhat.com/show_bug.cgi?id=1654217, link to redmine https://projects.theforeman.org/issues/25566

PR #303, opened 2018-11-14
Ensure we don't end up with plans stuck in planning

adamruzicka:
If the process planning an execution plan is killed before it can finish,
there's a chance it will get stuck in this state forever. This commit introduces
planning locks so we can stop the execution plans which are not really being
planned anymore during invalidation.

### DIFF
--- a/lib/dynflow/coordinator.rb
+++ b/lib/dynflow/coordinator.rb
@@ -294,6 +294,22 @@ module Dynflow
       end
     end
 
+    class PlanningLock < LockByWorld
+      def initialize(world, execution_plan_id)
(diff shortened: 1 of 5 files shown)


adamruzicka:
Pending tests, reproducer example here https://gist.github.com/adamruzicka/51f1a65deb7e46c8d25d13c8fa926f22

adamruzicka:
Rebased

iNecas:
Tested and works well also with Foreman-related projects. Could you think of a way how to add some test to cover the issue?

adamruzicka:
Added some tests and they're green

PR #305, opened 2018-11-19
Copy the favicon from doc to web assets

adamruzicka:
We need the logo in two places, in docs and in web/assets. Our solution was to have it in docs and the other location be just a symlink to the one in docs. In packaging we then dropped the docs part and the gem was left without a favicon, because it had only a broken symlink instead of the logo. Swapping the real file would just move the issue to the other part. Copying the actual icon seems to fix the issue. This could be also fixed in packaging but that seemed more error-prone to me.


iNecas:
Works for me. Thanks @adamruzicka

iNecas:
Cross-linking the redmine issue https://projects.theforeman.org/issues/21376

PR #307, opened 2018-11-23
Introduce post init hooks

adamruzicka:


lib/dynflow/rails/configuration.rb
```diff
@@ -65,6 +66,14 @@ def run_on_init_hooks(executor, world)
         source.each { |init| init.call(world) }
       end
 
+      def post_init(&block)
```
iNecas: Given this is for executor only, I guess we should call it 'post_executor_init`, or something like that, to be more explicit on when it's expected to run.

iNecas:
Thanks @adamruzicka

PR #309, opened 2018-11-27
Pin concurrent-ruby gem

jlsherrill:


adamruzicka:
Why `1.0.3` specifically?

jlsherrill:
Sorry, i meant to put an explanation, but i got pulled to a couple of meetings!  It seems dynflow uses Concurrent::Edge::Future which got removed in: https://github.com/ruby-concurrency/concurrent-ruby/pull/522
I'm a little confused as to why we are just now hitting the issue, as this is quite old.  

The initial problem was after pulling latest katello and foreman and bundle installing, we started getting this error:
```
LoadError: cannot load such file -- concurrent/lazy_register
/home/vagrant/.rvm/gems/ruby-2.5.1/gems/bootsnap-1.3.2/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require'
/home/vagrant/.rvm/gems/ruby-2.5.1/gems/activesupport-5.2.1/lib/active_support/dependencies.rb:287:in `block in require'
/home/vagrant/.rvm/gems/ruby-2.5.1/gems/activesupport-5.2.1/lib/active_support/dependencies.rb:253:in `load_dependency'
/home/vagrant/.rvm/gems/ruby-2.5.1/gems/activesupport-5.2.1/lib/active_support/dependencies.rb:287:in `require'
/home/vagrant/.rvm/gems/ruby-2.5.1/gems/concurrent-ruby-edge-0.2.0/lib/concurrent-edge.rb:7:in `<main>'
```

I then noticed that the version of c-r-edge was very old (0.2.4) compared to the version of c-r (1.1.3). These should  be 'locked' together (not the same version, but 1.1.3 should require 0.4.1).  I commented out the version specifier for c-r-edge, since the requirements in those two gems should keep them in sync, and then i started getting:

```
/home/vagrant/.rvm/gems/ruby-2.5.1/gems/bootsnap-1.3.2/lib/bootsnap/load_path_cache/core_ext/active_support.rb:74:in `block in load_missing_constant': uninitialized constant Concurrent::Edge::Future (NameError)
	from /home/vagrant/.rvm/gems/ruby-2.5.1/gems/bootsnap-1.3.2/lib/bootsnap/load_path_cache/core_ext/active_support.rb:8:in `without_bootsnap_cache'
	from /home/vagrant/.rvm/gems/ruby-2.5.1/gems/bootsnap-1.3.2/lib/bootsnap/load_path_cache/core_ext/active_support.rb:74:in `rescue in load_missing_constant'
	from /home/vagrant/.rvm/gems/ruby-2.5.1/gems/bootsnap-1.3.2/lib/bootsnap/load_path_cache/core_ext/active_support.rb:56:in `load_missing_constant'
	from /home/vagrant/git/dynflow/lib/dynflow/director.rb:16:in `block in <class:Director>'
```
which seems to have been removed in .  

I do wonder if the way that concurrent-ruby-edge is defining the version requirements from a file located in concurrent-ruby (lib/concurrent-ruby/version.rb), is causing weirdness, because i'm seeing versions get installed together which should not be.   I'm also having difficulty getting back into a broken state after this fix is in place (i have to empty the gemset before it will 'upgrade' the gems).

jlsherrill:
I'm also very confused as to why this is just now occurring, so please let me know if i'm missing something :)

jlsherrill:
Although i guess < 1.1.0 should be fine, let me update to that

adamruzicka:
We have a patch[1] ready to make Dynflow compatible with the latest and greatest concurrent-ruby, it is just waiting to be merged. So we just need to make this somehow work until that patch gets in.

[1] - https://github.com/Dynflow/dynflow/pull/296

iNecas:
Thanks @jlsherrill

iNecas:
Released in dynflow-1.1.3

PR #310, opened 2018-11-28
 Replace SortedSet with PriorityQueue in Clock 

adamruzicka:
Clock should now scale linearly with the number of events.

```
# bundle exec ruby examples/clock_benchmark.rb       
       user     system      total        real
   100  0.009689   0.003122   0.012811 (  0.022869)
  1000  0.092639   0.011817   0.104456 (  0.096092)
  5000  0.389363   0.040948   0.430311 (  0.399498)
 10000  1.041338   0.084084   1.125422 (  1.023096)
 50000  5.399935   0.637752   6.037687 (  5.478751)
100000 12.343072   1.202518  13.545590 ( 12.339196)
```

PR #311, opened 2018-12-10
Implement Clock using concurrent-ruby primitives

adamruzicka:
Based on https://github.com/Dynflow/dynflow/pull/310

### DIFF
--- a/lib/dynflow/clock.rb
+++ b/lib/dynflow/clock.rb
@@ -3,7 +3,6 @@ module Dynflow
 
     include Algebrick::Types
 
-    Tick  = Algebrick.atom
     Timer = Algebrick.type do
       fields! who:   Object, # to ping back
               when:  Time, # to deliver
@@ -40,19 +39,9 @@ module Dynflow
       end
     end
 
-    Pills = Algebrick.type do
-      variants None = atom,
-               Took = atom,
-               Pill = type { fields Float }
-    end
-
     def initialize(logger = nil)
-      @logger        = logger
-      @timers        = Utils::PriorityQueue.new { |a, b| b <=> a }
-      @sleeping_pill = None
-      @sleep_barrier = Mutex.new
-      @sleeper       = Thread.new { sleeping }
-      Thread.pass until @sleep_barrier.locked? || @sleeper.status == 'sleep'
+      @logger = logger
+      @timers = Utils::PriorityQueue.new { |a, b| b <=> a }
     end
 
     def default_reference_class
@@ -60,9 +49,7 @@ module Dynflow
     end
 
     def on_event(event)
-      if event == :terminated
-        @sleeper.kill
-      end
+      wakeup if event == :terminated
     end
 
     def tick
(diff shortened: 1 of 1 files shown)


lib/dynflow/clock.rb
```diff
@@ -84,43 +70,31 @@ def add_timer(timer)
...
-      @sleep_barrier.synchronize do
-        loop do
-          @sleeping_pill = None
-          @sleep_barrier.sleep
-          pill           = @sleeping_pill
-          @sleeping_pill = Took
-          @sleep_barrier.sleep pill.value
-          reference.tell(:tick)
-        end
+      @timer = Concurrent::Channel.timer(sec)
```
iNecas: Any reason for using Channel vs. [ScheduledTask](http://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/ScheduledTask.html)?

adamruzicka: We may need to wake the timer earlier than it was scheduled to, `Channel.timer` provides nicer interface for that

iNecas:
Looks good code-wise: I've tried to run the https://github.com/Dynflow/dynflow-benchmark tool against it, but it fails with:

```
be ../dynflow-benchmark/benchmark.rb -d 0.001 -i 0.01 -m 10 -p 200
/home/inecas/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/concurrent-ruby-1.1.4/lib/concurrent/synchronization/mutex_lockable_object.rb:16: [BUG] Segmentation fault at 0x000055dd39332fab
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
```

I've tried to move some dynflow stuff into separate fork https://github.com/Dynflow/dynflow-benchmark/tree/prepare-in-fork, but it still occasionally failing (not that much though).

The times are similar with the new implementation or the old one. I think we should figure out what's wrong before merging though. Seems like some concurrent ruby primitives are not cleaned after forking, and are causing issues.

adamruzicka:
Opened https://github.com/Dynflow/dynflow-benchmark/pull/1 PR based on 
[Dynflow/dynflow-benchmark@20eaf22](https://github.com/Dynflow/dynflow-benchmark/commit/20eaf2257463d1fb3f0f4e092b190a7a1abd3eed) to handle the failures in dynflow-benchmark

adamruzicka:
Test failures are being handled in https://github.com/Dynflow/dynflow/pull/321

lib/dynflow/clock.rb
```diff
@@ -84,18 +84,17 @@ def first_timer
...
     def sleep_to(timer)
       return unless timer
       sec = [timer.when - Time.now, 0.0].max
-      @timer = Concurrent::Channel.timer(sec)
-      Concurrent::Channel.go do
-        ~@timer
-        return if terminating?
-        reference.tell(:tick)
-      end
+      @timer = Concurrent::ScheduledTask.execute(sec) { reference.tell(:tick) unless terminating? }
```
ezr-ondrej: [Docs](https://github.com/ruby-concurrency/concurrent-ruby/blob/d11b29c37b81320ca7126cc9cd85f4f3d17a78a3/lib/concurrent/scheduled_task.rb#L22) are saying sec == 0 is going to fail. I haven't tested it though.

adamruzicka: Docs aren't completely right
```
> Concurrent::ScheduledTask.execute(-1) { puts "boom" }; nil
ArgumentError: seconds must be greater than zero
from /home/aruzicka/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/concurrent-ruby-1.1.5/lib/concurrent/scheduled_task.rb:167:in `initialize'

> Concurrent::ScheduledTask.execute(0) { puts "boom" }; nil
=> nil
boom
```

will fix anyway

adamruzicka:
Rebased and squashed the commits.

ezr-ondrej:
AWT :)

adamruzicka:
One of the tests failed with this, maybe we should relax the requirement a bit? Not sure if we can make the clock more precise somehow

```
Dynflow::Clock#test_0003_pongs on expected times [/home/travis/build/Dynflow/dynflow/test/clock_test.rb:34]:
Expected |0.1 - 0.642421603| (0.542421603) to be <= 0.08.
```

adamruzicka:
I'll rekick the entire test suite a couple of times to see if it was just travis being busy or if it is a thing we should handle.

ezr-ondrej:
You think `ScheduledTask` can be less precise?

adamruzicka:
I don't think so, but I'd be glad if we could avoid transient test failures

ezr-ondrej:
Yeah, that would be nice... but with the time it's never 100% but this seems like solid :green_apple: percentage

adamruzicka:
It seems stable enough

PR #312, opened 2018-12-11
Enhance fix for termination with running action

iNecas:
Previously, we would not respond to the pool in case of any exception
coming from the `work_item.execute`. Even though I don't think we expect
any other exceptions other than PeristenceError, it didn't feel right not to
respond on any exception that might occur.

When I looked at API of concurrent ruby, it seems like we can use check on
`global_io_executor` to check it's still operational.


adamruzicka:
Thanks @iNecas !

PR #313, opened 2018-12-12
Make sure an execution plan cannot be its own sub plan

adamruzicka:
Previously the following would hold as a definition of a sub plan:
Execution plan E2 is a sub plan of E1 if E2 has at least one action with
caller_execution_plan_id set to E1.id.

Let's consider an execution plan E1 with a single action A1. A1's
caller_{execution_plan,action}_id is nil and everything is in order.

However if we would plan action A2 from A1, A2's caller_execution_plan_id would
be set to E1.id and caller_action_id to A1.id. According to the previous
definition E1 would be considered a sub plan of itself.

With this patch only execution plan's entry action can have
caller_execution_plan_id set. This way, we shouldn't reach a situation where an
execution plan is its own sub plan.

### DIFF
--- a/lib/dynflow/action.rb
+++ b/lib/dynflow/action.rb
@@ -161,11 +161,11 @@ module Dynflow
     end
 
     def caller_action
-      plase! Present
+      phase! Present
       return nil if @caller_action_id
       return @caller_action if @caller_action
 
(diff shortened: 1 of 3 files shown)


iNecas:
Thanks @adamruzicka

PR #314, opened 2019-01-03
Fix travis on ruby 2.2

iNecas:
This fixes the following issue:

```
 The program 'bundle' is currently not installed. To run
'bundle' please ask your administrator to install the package 'bundler'.

```

Seems like the issue is the bundler 2.0 is not compatible with
ruby < 2.3.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ sudo: false
 language:
   - ruby
 
+before_install: >-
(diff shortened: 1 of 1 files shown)


adamruzicka:
ACK

adamruzicka:
Thanks @iNecas !

PR #315, opened 2019-01-24
Refs #25736 - add support for clearing Logging.mdc if present

iNecas:
Also, remove the `action_logger` phase constraint.


adamruzicka:
Makes sense to me, waiting for travis

PR #317, opened 2019-02-21
Make the job_provider_id available in the worker

iNecas:
Also make execution_plan.id match active_job.id, which makes it easier to
use single uuid to search the data across the active job and dynflow realm.

As part of this change, we've made it possible to set the execution plan id
to custom value (the developer needs to be careful to not use duplicate keys: with great power
comes great responsiblity).

### DIFF
--- a/test/execution_plan_test.rb
+++ b/test/execution_plan_test.rb
@@ -191,6 +191,18 @@ module Dynflow
         end
       end
 
+      describe 'custom plan id' do
+        let :execution_plan do
+          world.plan_with_options(action_class: Support::CodeWorkflowExample::IncomingIssues,
+                                  args: [issues_data],
+                                  id: 'my-unique-id')
+        end
+
+        it 'allows setting custom id for the execution plan' do
+          execution_plan.id.must_equal 'my-unique-id'
+        end
+      end
+
       describe 'planning algorithm' do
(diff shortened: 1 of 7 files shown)


iNecas:
@ezr-ondrej this should make your life easier

test/execution_plan_test.rb
```diff
@@ -191,6 +191,18 @@ module ExecutionPlanTest
         end
       end
 
+      describe 'custom plan id' do
+        let :execution_plan do
+          world.plan_with_options(action_class: Support::CodeWorkflowExample::IncomingIssues,
+                                  args: [issues_data],
+                                  id: 'my-unique-id')
```
adamruzicka: If I recall on postgres we use an `UUID` typed column for storing the id. Passing something what is not an uuid should fail. Any idea why that doesn't happen?

iNecas: Updated to use uuid. It turns out we were not using uuid for postgres due to incorrect database condition. See https://github.com/Dynflow/dynflow/pull/319 for more details.

iNecas:
Comments addressed, we're :green_heart: still

adamruzicka:
Thanks @iNecas & @ezr-ondrej !

PR #318, opened 2019-02-22
Reduce status updates on rescue

iNecas:
The previous rescue algorithm put the task into paused state and then we
looked at what are the possible rescue strategies, and if it was to
skip, we marked the steps to skip and executed the task again.

Several problems where there with this approach:

- the execution plan needed to get back at the beginning of the workers
  queue, so there was additional overhead with this

- the execution history got confusing, because there was note about
  finishing and starting the execution again

- it's hard to hook events (such as logging or notifications) for paused
  tasks, as when the plan was getting into paused state, we didn't know
  it would be rescued immediately.

This is a prerequisite for allowing notifications on paused tasks.

### DIFF
--- a/lib/dynflow/director.rb
+++ b/lib/dynflow/director.rb
@@ -144,24 +144,38 @@ module Dynflow
     def unless_done(manager, work_items)
       return [] unless manager
       if manager.done?
-        finish_manager(manager)
-        return []
+        try_to_rescue(manager) || finish_manager(manager)
       else
         return work_items
       end
     end
 
+    def try_to_rescue(manager)
+      rescue!(manager) if rescue?(manager)
+    end
+
     def finish_manager(manager)
-      @execution_plan_managers.delete(manager.execution_plan.id)
-      if rescue?(manager)
-        rescue!(manager)
-      else
-        set_future(manager)
+      execution_plan = manager.execution_plan
+      if execution_plan.state == :running
+        if execution_plan.error?
+          execution_plan.update_state(:paused)
+          execution_plan.execution_history.add('pause execution', @world.id)
+        elsif manager.done?
+          execution_plan.execution_history.add('finish execution', @world.id)
+          execution_plan.update_state(:stopped)
+        end
+        # If the state is marked as running without errors but manager is not done,
+        # we let the invalidation procedure to handle re-execution on other executor
       end
+      execution_plan.save
+      return []
+    ensure
+      @execution_plan_managers.delete(execution_plan.id)
+      set_future(manager)
     end
 
     def rescue?(manager)
-      if @world.terminating? || !(@world.auto_rescue && manager.execution_plan.state == :paused)
+      if @world.terminating? || !(@world.auto_rescue && manager.execution_plan.error?)
         false
       elsif !@rescued_steps.key?(manager.execution_plan.id)
         # we have not rescued this plan yet
@@ -179,11 +193,12 @@ module Dynflow
       # to put this logic of making sure we don't run rescues in endless loop
       @rescued_steps[manager.execution_plan.id] ||= Set.new
       @rescued_steps[manager.execution_plan.id].merge(manager.execution_plan.failed_steps.map(&:id))
-      rescue_plan_id = manager.execution_plan.rescue_plan_id
-      if rescue_plan_id
-        @world.executor.execute(rescue_plan_id, manager.future, false)
+      new_state = manager.execution_plan.prepare_for_rescue
+      if new_state == :running
+        return manager.restart
       else
-        set_future(manager)
+        manager.execution_plan.update_state(new_state)
+        return false
       end
     end
 
--- a/lib/dynflow/director.rb
+++ b/lib/dynflow/director.rb
@@ -119,9 +119,9 @@ module Dynflow
     def work_failed(work)
       if (manager = @execution_plan_managers[work.execution_plan_id])
         manager.terminate
-        # work failed means there was probably a database issue (execution plan gone missing). Don't try
-        # to store the status in that case.
-        finish_manager(manager, store: false)
+        # Don't try to store when the execution plan went missing
+        plan_missing = @world.persistence.find_execution_plans(:filters => { uuid: work.execution_plan_id }).empty?
+        finish_manager(manager, store: !plan_missing)
       end
     end
 
--- a/lib/dynflow/director.rb
+++ b/lib/dynflow/director.rb
@@ -216,8 +216,8 @@ module Dynflow
       execution_plan = manager.execution_plan
       if execution_plan.state == :running
         if execution_plan.error?
-          execution_plan.update_state(:paused)
           execution_plan.execution_history.add('pause execution', @world.id)
+          execution_plan.update_state(:paused)
         elsif manager.done?
           execution_plan.execution_history.add('finish execution', @world.id)
           execution_plan.update_state(:stopped)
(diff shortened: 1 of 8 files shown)


lib/dynflow/execution_plan.rb
```diff
@@ -199,16 +199,17 @@ def sub_plans_count
       persistence.find_execution_plan_counts(filters: { 'caller_execution_plan_id' => self.id })
     end
 
-    def rescue_plan_id
+    def prepare_for_rescue
```
iNecas: Could not think of better name

lib/dynflow/execution_plan.rb
```diff
@@ -232,14 +233,6 @@ def steps_in_state(*states)
       self.steps.values.find_all {|step| states.include?(step.state) }
     end
 
-    def rescue_from_error
```
iNecas: This method has only be called from tests

test/abnormal_states_recovery_test.rb
```diff
@@ -198,7 +198,7 @@ def with_invalidation_while_executing(finish)
           plan = client_world.persistence.load_execution_plan(triggered.id)
           plan.state.must_equal :paused
           expected_history = [['start execution', executor_world.id],
-                              ['finish execution', executor_world.id]]
+                              ['pause execution', executor_world.id]]
```
iNecas: Added a new kind of execution history for paused event so that it's more descriptive on what happened.

test/rescue_test.rb
```diff
@@ -15,125 +15,99 @@ def execute(*args)
       end
 
       let :rescued_plan do
-        execution_plan.rescue_from_error.value
+        world.persistence.load_execution_plan(execution_plan.id)
```
iNecas: Instead of calling the method that is being used only in tests, we look at the current state of execution plan after the execution finished

test/rescue_test.rb
```diff
@@ -15,125 +15,99 @@ def execute(*args)
...
 
-        it 'suggests skipping the action' do
-          execution_plan.rescue_strategy.must_equal Action::Rescue::Skip
-        end
+          it 'suggests skipping the action' do
+            execution_plan.rescue_strategy.must_equal Action::Rescue::Skip
+          end
 
-        it 'skips the action and continues' do
-          rescued_plan.state.must_equal :stopped
```
iNecas: By default, the world in tests has `auto_rescue` set to false: therefore it doesn't rescue at all: all the auto-rescuing tests were moved to the `auto rescue` block of the test

test/test_helper.rb
```diff
@@ -225,11 +225,11 @@ def finish_the_plan(triggered)
   def assert_plan_reexecuted(plan)
     assert_equal :stopped, plan.state
     assert_equal :success, plan.result
-    assert_equal plan.execution_history.map(&:name),
-                 ['start execution',
+    assert_equal ['start execution',
```
iNecas: Just swapping the order: the expected value should be first

test/abnormal_states_recovery_test.rb
```diff
@@ -142,13 +142,7 @@ def with_invalidation_while_executing(finish)
...
 
         before do
-          client_world.persistence.find_execution_plans({}).each do |plan|
-            # make sure we don't handle plans from previous tests
-            # TODO: delete the plans instead, once we have
-            # https://github.com/Dynflow/dynflow/pull/141 merged
-            plan.set_state(:stopped, true)
-            plan.save
-          end
+          client_world.persistence.delete_execution_plans({})
```
iNecas: Took us time to fix this TODO, but here we are

lib/dynflow/director.rb
```diff
@@ -144,24 +146,26 @@ def terminate
...
-        rescue!(manager)
-      else
-        set_future(manager)
-      end
+      set_future(manager)
     end
 
     def rescue?(manager)
-      if @world.terminating? || !(@world.auto_rescue && manager.execution_plan.state == :paused)
+      if @world.terminating? || !(@world.auto_rescue && manager.execution_plan.error?)
```
iNecas: We needed to do this, otherwise the plan in paused/pending would get rescued as well

test/executor_test.rb
```diff
@@ -138,6 +138,7 @@ def plan
             wait_for('execution plan removed from executor') do
               !director.current_execution_plan_ids.include?(execution_plan.id)
             end
+            world.persistence.find_execution_plans(filters: { uuid: [execution_plan.id] }).must_be :empty?
```
iNecas: I was running into the issue that the procedure on failure was still trying to update the state of the execution plan (that it still had in memory), so it saved it with invalid state, and was causing issues in other parts: this way we make sure the execution plan doesn't re-appear.

iNecas:
Everything I've planned on doing in this PR, I've done

lib/dynflow/execution_plan.rb
```diff
@@ -199,16 +199,17 @@ def sub_plans_count
...
       when Action::Rescue::Fail
-        update_state :stopped
-        nil
+        return :stopped
       when Action::Rescue::Skip
         failed_steps.each { |step| self.skip(step) }
-        self.id
+        return :running
+      else
+        return :paused
```
adamruzicka: None of these returns is really needed, no need to change, just saying.

lib/dynflow/director.rb
```diff
@@ -119,7 +119,9 @@ def work_finished(work)
     def work_failed(work)
       if (manager = @execution_plan_managers[work.execution_plan_id])
         manager.terminate
-        finish_manager(manager)
+        # work failed means there was probably a database issue (execution plan gone missing). Don't try
```
adamruzicka: What are the other cases when this may happen? Can we safely ignore those?

lib/dynflow/director.rb
```diff
@@ -207,6 +212,22 @@ def track_execution_plan(execution_plan_id, finished)
...
+          execution_plan.update_state(:paused)
+          execution_plan.execution_history.add('pause execution', @world.id)
+        elsif manager.done?
+          execution_plan.execution_history.add('finish execution', @world.id)
+          execution_plan.update_state(:stopped)
+        end
+        # If the state is marked as running without errors but manager is not done,
+        # we let the invalidation procedure to handle re-execution on other executor
+      end
+      execution_plan.save
```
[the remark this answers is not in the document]
iNecas: Good catch. Should not be needed. Updated

lib/dynflow/execution_plan.rb
iNecas: updated

lib/dynflow/director.rb
iNecas: In can't think of any other one, but just in case, I added additional check for presence of the execution plan: this should not happen then often, so some additional check to decide whether to store or not make sense.

iNecas:
All comments addressed

iNecas:
@adamruzicka nice catch regarding the execution history. Should be fixed now (+ added assertions into test to ensure we will not miss it next time)

lib/dynflow/director.rb
```diff
@@ -207,6 +212,25 @@ def track_execution_plan(execution_plan_id, finished)
...
+          execution_plan.execution_history.add('pause execution', @world.id)
+          execution_plan.update_state(:paused)
+        elsif manager.done?
+          execution_plan.execution_history.add('finish execution', @world.id)
+          execution_plan.update_state(:stopped)
+        end
+        # If the state is marked as running without errors but manager is not done,
+        # we let the invalidation procedure to handle re-execution on other executor
+      when :paused, :stopped
+        execution_plan.execution_history.add('finish execution', @world.id)
```
adamruzicka: Shouldn't it be `pause execution` when the execution plan gets paused?

iNecas: good catch. updated

adamruzicka:
Thanks @iNecas !

PR #323, opened 2019-03-19
Support for logger message with block

iNecas:
Before this patch, we couldn't do things like this:

    logger.debug { "Message that needs some calculation" }


adamruzicka:
Thanks @iNecas !

PR #324, opened 2019-03-22
Fix pause hooks + refactor history notices

iNecas:
After 8bae5b62b273, we had a code that was updating `state` outside of
`update_state` method, which meant the hooks (observed with `:paused`
state) were not run when `auto_rescue` was enabled.

I've also noticed coupling between adding history and `update_state`
methods, so I put it directly into `update_state`.


adamruzicka:
Thanks @iNecas !

PR #325, opened 2019-04-11
Fix root_action for post-planning phase

iNecas:
Before this patch, the root_action was returning correctly only in
planning phase.

Checking for @caller_action_id.nil? or @caller_execution_plan_id being
set (which means the the caller_action_id is coming from different
execution plan id) should extend this functionality for other phases as
well.

### DIFF
--- a/lib/dynflow/action.rb
+++ b/lib/dynflow/action.rb
@@ -571,7 +571,15 @@ module Dynflow
     end
 
     def root_action?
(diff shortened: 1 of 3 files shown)


adamruzicka:
Thanks @iNecas !

iNecas:
Thanks for quick merge: released dynflow 1.2.3 with this change

PR #327, opened 2019-06-25
Fixes #27123 - Make "force_kill_waittime" option adjustable

hao-yu:
Also set this option to default 60 seconds so that long running
task is more likely to finish before being killed.


iNecas:
Thanks @hao-yu

PR #329, opened 2019-07-17
Add support for running workers via sidekiq

iNecas:
This PR adds support for running workers via sidekiq, while preserving the current executor
for cases where sidekiq is not around.

Things to add within this PR:

- [x] - telemetry support for sidekiq workers
- [x] - unit tests

Things to be covered after this PR:

There are still many things that need to be covered before turning into production. However
it should be good enough to start playing with and I would prefer addressing the additional
steps in separate PRs once this is merged

Specific steps TBD.

See the `examples/remote_executor.rb` for instructions how to test it out.

This PR is split into smaller comments that should make sense by them own. The recommended approach to review the PR is going though commits one by one (the commit messages also have some additional information about the reasons for the changes)

EDIT:
Redmine: https://projects.theforeman.org/issues/27409


iNecas:
Definitely

lib/dynflow/director.rb
```diff
@@ -42,33 +64,70 @@ def initialize(execution_plan_id, step, queue)
...
+        manager = Director::SequentialManager.new(world, execution_plan)
+        manager.finalize
+        @finalize_steps_data = manager.finalize_steps.map(&:to_hash)
+      end
+
+      def to_hash
+        super.merge(finalize_steps_data: @finalize_steps_data)
+      end
+
+      def self.new_from_hash_(hash, *_args)
```
adamruzicka: To fix execution plans with finalize steps ending in stopped-pending do `sed 's/_(/(/'` on this line

iNecas: Sorry, my bad: I was just testing that the code is not working without the method (and you've noticed it wasn't), and forgot to put it back. Updated

iNecas:
It seems I've found the fix for the mysql failures: seems like it had something to do with the way the rounding of time is handled in specific mysql version: uncreasing the datetime precision for mysql seemed to put the mysql test failure away.

iNecas:
It seems I've found the fix for the mysql failures: seems like it had something to do with the way the rounding of time is handled in specific mysql version: uncreasing the datetime precision for mysql seemed to put the mysql test failure away.

iNecas:
@adamruzicka thinking about the sidekiq redis settings, I'm actually not sure it belongs to the dynflow configuration, as it's something to be setup globally when running the processes. It might be even bigger issue when running in an environment where sidekiq is already setup, and one would like just to use it. Therefore I would probably suggest leaving the sidekiq/redis configuration outside of Dynflow itself, and Dynflow just assuming the configuration is correct.

Are there any downsides that I'm overlooking?

test/abnormal_states_recovery_test.rb
```diff
@@ -275,7 +275,7 @@ def with_invalidation_while_executing(finish)
             client_world_config = Config::ForWorld.new(Config.new.tap { |c| c.executor = false }, create_world )
             client_world_config.auto_validity_check.must_equal false
 
-            executor_world_config = Config::ForWorld.new(Config.new.tap { |c| c.executor = lambda { |w, _| Executors::Parallel.new(w, 15) } }, create_world )
+            executor_world_config = Config::ForWorld.new(Config.new.tap { |c| c.executor = lambda { |w, _| Executors::Parallel::Core } }, create_world )
```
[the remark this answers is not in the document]
iNecas: Updated

examples/example_helper.rb
```diff
@@ -41,15 +46,26 @@ def persistence_adapter
     end
 
     def logger_adapter
-      Dynflow::LoggerAdapters::Simple.new $stderr, 4
+      Dynflow::LoggerAdapters::Simple.new $stderr, 1
```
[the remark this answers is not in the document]
iNecas: Updated

lib/dynflow/executors/abstract/core.rb
```diff
@@ -0,0 +1,124 @@
...
+        end
+
+        def schedule_heartbeat
+          @world.clock.ping(self, @heartbeat_interval, :heartbeat)
+        end
+
+        def on_message(message)
+          super
+        rescue Errors::PersistenceError => e
+          self.tell([:handle_persistence_error, e])
```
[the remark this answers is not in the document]
iNecas: Updated

iNecas:
Telemetry support added for the sidekiq

test/executor_test.rb
```diff
@@ -2,444 +2,455 @@
...
+require 'sidekiq/testing'
+::Sidekiq::Testing::inline!
+require 'dynflow/executors/sidekiq/core'
+
 module Dynflow
   module ExecutorTest
-    describe "executor" do
-
-      include PlanAssertions
+    [::Dynflow::Executors::Parallel::Core, ::Dynflow::Executors::Sidekiq::Core].each do |executor|
```
iNecas: I've added both executors to test, only line 17-19 (cleanup of the process_word), 23 (setting the executor) and 24 (setting the process role) has been modified. The rest is just whitespace

iNecas:
Starting with the `telemetry` commit, we've started hitting this error with postgres:

```
consistency check::for plans assigned to invalid world::world invalidation#test_0002_schedules the plans to be run on different executor:
Sequel::PoolTimeout: timeout: 5.0, elapsed: 5.000237526000035
```
I haven't seen this before (and test tests just one commit sooner were passing fine). I'm also failing to reproduce this locally :(

iNecas:
I was doing some experiments https://github.com/Dynflow/dynflow/pull/330 with the commits around the failure and eventually things got green on multiple different versions (with or without telemetry: not that I would be expect the results should be different).

Last 4 travis runs all got green (counting 3 recent in the experimental PR + 1 last here), so I still hope it's been related to some specific error in Travis: I was running the tests in a loop at my side and none of those issues have showed up.

Also, I'm glad the test coverage we have now would have caught the `new_from_hash_` mistake I did before. This PR should be ready for final review.

iNecas:
Just to keep some organization around, I've filed an redmine issue for this PR https://projects.theforeman.org/issues/27409

lib/dynflow/config.rb
```diff
@@ -97,8 +97,28 @@ def queues
...
+
+    def validate_executor!(value)
+      accepted_executors = [Executors::Parallel::Core]
+      accepted_executors << Executors::Sidekiq::Core if defined? Executors::Sidekiq::Core
+      if value && !accepted_executors.include?(value)
+        raise ArgumentError, "Executor #{value} is expected to be one of #{accepted_executors.inspect}"
+      end
+    end
+
+    # does the work represent some process-wide role (important for sikekiq-based deployments)
```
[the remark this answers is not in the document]
iNecas: Updated. Force-pushed the change directly to the commit that introduce it. Nothing else has changed

iNecas:
Everything is still :green_apple:

lib/dynflow/executors/sidekiq/worker_jobs.rb
```diff
@@ -0,0 +1,34 @@
...
+            with_telemetry(work_item) do
+              Executors.run_user_code do
+                work_item.world = Dynflow.process_world
+                work_item.execute
+              end
+            end
+          rescue Errors::PersistenceError => e
+            OrchestratorJobs::HandlePersistenceError.perform_async(e, work_item)
+          ensure
+            OrchestratorJobs::WorkerDone.perform_async(work_item)
```
adamruzicka: When a persistence error happens, both jobs get queued. However on orchestrator's side, the first one which arrives (`HandlePersistenceError`) removes the execution plan from all the internal structures and then the `WorkerDone` is just a no-op. Do we really need to send both?

When an `Errors::FatalPersistenceError` is raised on the worker, it gets caught and sent back to the orchestrator. The orchestrator reacts to this by terminating. There are two things wrong with this: the worker which had the error keeps running and the orchestrator which may be working fine terminates.

adamruzicka: And what's even worse is the orchestrator process does not die, just the world inside it terminates

lib/dynflow/config.rb
```diff
@@ -97,8 +97,28 @@ def queues
...
+    def validate_executor!(value)
+      accepted_executors = [Executors::Parallel::Core]
+      accepted_executors << Executors::Sidekiq::Core if defined? Executors::Sidekiq::Core
+      if value && !accepted_executors.include?(value)
+        raise ArgumentError, "Executor #{value} is expected to be one of #{accepted_executors.inspect}"
+      end
+    end
+
+    # does the work represent some process-wide role (important for sidekiq-based deployments)
+    config_attr :process_role do |world, config|
```
ezr-ondrej: Why is this introduced? Why do we need to separate the worlds? I do not see any particular reason sidekiq world can't be shared if required. This could lead to completly different flows for sidekiq and without, if we start branch the code depending on this (and if we won't why we need it).
If we are at it, maybe we could do it for all worlds, not just for sidekiq?

adamruzicka:
Big thank you @iNecas !

PR #334, opened 2019-08-14
Bump rubocop ruby parser version to 2.3

adamruzicka:


test/support/middleware_example.rb
```diff
@@ -99,7 +100,7 @@ def run
       end
 
       def filter_sensitive_data
-        output[:spell] = '***'
+        output[:spell] = '***'.dup
```
ezr-ondrej: Should I want to know why? xD

ezr-ondrej:
The dups are going to be necessary all around in Ruby 3.0? That's bit sad :thinking:

test/support/middleware_example.rb
adamruzicka: We do some modifications of these strings using `<<`. By adding the magic comment (`# frozen_string_literal: true`) all string literals become frozen by default and exceptions are raised when we try to modify them. `dup`ing a frozen string creates an un-frozen copy of it, which can be modified.

ezr-ondrej:
And are we sure we `dup`ed enough? :rofl:

ezr-ondrej:
I mean are we confident the test discovered all the frozen strings, that shouldn't have been frozen?

adamruzicka:
Pretty sure, as far as I'm aware we don't use, let alone modify strings much.

PR #335, opened 2019-08-14
 Allow running multiple orchestrators in active-passive mode

adamruzicka:
Upon startup, the orchestrator checks if dynflow_orchestrator_uuid key is unset in Redis. If so, it acquires an orchestrator lock by associating the uuid of the orchestrator's world with the key. However this lock has limited time to live and needs to be periodically
refreshed.

If the key is set, the orchestrator goes into passive mode and periodically polls the lock's status. If it manages to acquire the lock, it becomes the active orchestrator.

Active orchestrator periodically updates the record in Redis to ensure its freshness.

Redmine: https://projects.theforeman.org/issues/27412

### DIFF
--- a/lib/dynflow/executors/sidekiq/redis_locking.rb
+++ b/lib/dynflow/executors/sidekiq/redis_locking.rb
@@ -1,0 +1,68 @@
+module Dynflow
+  module Executors
+    module Sidekiq
+      module RedisLocking
+        REDIS_LOCK_KEY = 'dynflow_orchestrator_uuid'
+        REDIS_LOCK_TTL = 60
+        REDIS_LOCK_POLL_INTERVAL = 15
+
+        ACQUIRE_OK = 0
+        ACQUIRE_MISSING = 1
+        ACQUIRE_TAKEN = 2
+
+        RELEASE_SCRIPT = <<~LUA
+          if redis.call("get", KEYS[1]) == ARGV[1] then
+            redis.call("del", KEYS[1])
+          end
+          return #{ACQUIRE_OK}
+        LUA
+
+        REACQUIRE_SCRIPT = <<~LUA
+          if redis.call("exists", KEYS[1]) == 1 then
+            local owner = redis.call("get", KEYS[1])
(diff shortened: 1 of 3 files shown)


adamruzicka:
Rubocop should be fixed by https://github.com/Dynflow/dynflow/pull/334

lib/dynflow/executors/sidekiq/redis_locking.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        def release_orchestrator_lock
+          ::Sidekiq.redis { |conn| conn.eval RELEASE_SCRIPT, [REDIS_LOCK_KEY], [@world.id] }
+        end
+
+        def wait_for_orchestrator_lock
+          mode = nil
+          loop do
+            active = ::Sidekiq.redis do |conn|
+              conn.set(REDIS_LOCK_KEY, @world.id, :ex => REDIS_LOCK_TTL, :nx => true)
```
ezr-ondrej: Is the set operation on redis atomic?

adamruzicka: Yes

adamruzicka:
I looked at it briefly and I didn't exactly set our use case. We need to take the lock, make it expire after a while, but keep refreshing it as long as we hold it. As far as I can tell, the refreshing part is not implemented there. But I'd be happy to be proven wrong :)

ezr-ondrej:
no I haven't seen the decay introduced there. It's effective only for a period of time, and thus can be probably forced to way we want, but it wouldn't feel right.
I'm just thinking if this actually make sense to run more orchestrators just to make the takeover faster opposed to restart of the orchestrator?
On the other hand it probably makes sense to lock the redis to prevent issues if multiple executors are ran by accident, right?

adamruzicka:
This was meant mainly as a safeguard to prevent accidentally running multiple orchestrators. I could change it to only prevent another orchestrator from starting if some is already active, but imo this gives a better user experience.

PR #337, opened 2019-08-16
Start MySQL service only in specific job

cotsog:
Also removed the deprecated `sudo:` key and fixed `language:`


adamruzicka:
Thank you @cotsog !

PR #339, opened 2019-09-10
update plan to stopped if zero steps during invalidation

jlsherrill:


jlsherrill:
you mentioned something about if 'ExecutionPlan#steps' queried the database everytime, and it does NOT appear like it does.  It looks like its just an in memory structure, i wasn't sure of the implication of that.

adamruzicka:
> it does NOT appear like it does.

Then it is alright. LGTM but rubocop is not happy

adamruzicka:
Thank you @jlsherrill !

PR #342, opened 2019-10-10
Dependencies for running as a service

adamruzicka:


lib/dynflow/rails/configuration.rb
```diff
@@ -32,6 +32,8 @@ class Configuration
       # the orchestration tied to the models.
       attr_accessor :disable_active_record_actions
 
+      attr_accessor :process_role
```
ezr-ondrej: I still do not like the name of this config, I feel it missleading. We want to be able to have one sidekiq process for worker and orchestrator together, right?

bin/dynflow-sidekiq.rb
```diff
@@ -0,0 +1,30 @@
...
+
+rails_env_file = File.expand_path('./config/environment.rb', rails_root)
+
+require rails_env_file
+role = nil
+if Sidekiq.options[:queues].include?("dynflow_orchestrator")
+  ::Rails.application.dynflow.executor!
+  role = :orchestrator
+elsif (Sidekiq.options[:queues] - ['dynflow_orchestrator']).any?
+  require 'sidekiq-reliable-fetch'
```
ezr-ondrej: This is redundant.

lib/dynflow/rails/configuration.rb
adamruzicka: Yeah, it is kinda misleading. If both are in the same process, which is totally doable, they are considered to be orchestrator, but still perform both duties.

adamruzicka:
Regarding docs, commit message of af4fb7b gives at least a hint, but a full-fledged documentation would be nice

ezr-ondrej:
Rubocop still complains btw.

bin/dynflow_sidekiq.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+rails_root = Dir.pwd
```
ezr-ondrej: Would it make sense to make this configurable? Maybe by a env variable?

ezr-ondrej:
And has a conflict :(

adamruzicka:
Resolved

extras/dynflow@.service
```diff
@@ -0,0 +1,13 @@
+# /usr/lib/systemd/system/dynflow@.service
+[Unit]
+Description=Dynflow sidekiq worker
+After=network.target remote-fs.target nss-lookup.target
+
+[Service]
+TimeoutSec=90
+ExecStart=/usr/bin/sh -c 'exec sidekiq -C /etc/dynflow/%i.yml $DYNFLOW_OPTIONS'
```
ezr-ondrej:
```suggestion
ExecStart=/usr/bin/sh -c 'bundle exec sidekiq -C /etc/dynflow/%i.yml $DYNFLOW_OPTIONS'
```
Or
```suggestion
ExecStart=/usr/bin/bundle exec sidekiq -C /etc/dynflow/%i.yml $DYNFLOW_OPTIONS
```
Or (and change env file accordingly)
```suggestion
ExecStart=/usr/bin/bundle exec sidekiq -e $RAILS_ENV -r PATH_TO_SCRIPT -C /etc/dynflow/%i.yml 
```

adamruzicka: The shell was there to make the env variables work. As far as I know systemd doesn't translate those, so that rules out numbers 2 and 3.

The exec was there to replace the shell process with the sidekiq one so we don't keep yet another process around. Also I'm not sure how good idea is to use `bundle` on systems where sidekiq is installed from rpms and uses system ruby.

ezr-ondrej: Me neighter, so let's not. Thanks for clarifying.

adamruzicka: Oh, so according to the docs[1, seventh paragraph] systemd can do env var substitution, just the behavior is slightly different from shell. That would mean I could drop the shell part

[1] - https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines

PR #344, opened 2019-10-18
Refs #27410 - use new API within action

ezr-ondrej:
Use the `plan_event` API for sending events to the action.

### DIFF
--- a/lib/dynflow/action.rb
+++ b/lib/dynflow/action.rb
@@ -89,6 +89,13 @@ module Dynflow
       end
     end
 
+    DelayedEvent = Algebrick.type do
+      fields! execution_plan_id: String,
+              step_id:           Integer,
+              event:             Object,
+              time:              type { variants Time, NilClass }
+    end
+
     def self.constantize(action_name)
       super action_name
     rescue NameError
@@ -326,9 +333,8 @@ module Dynflow
     # Plan an +event+ to be send to the action defined by +action+, what defaults to be self.
     # if +time+ is not passed, event is sent as soon as possible.
     def plan_event(event, time = nil, execution_plan_id: self.execution_plan_id, step_id: self.run_step_id)
-      Type! time, Time, Numeric, NilClass
-      time  = Time.now + time if time.is_a? Numeric
-      delayed_events << Dispatcher::Event[execution_plan_id, step_id, event, time]
+      time = @world.clock.current_time + time if time.is_a?(Numeric)
+      delayed_events << DelayedEvent[execution_plan_id, step_id, event, time]
     end
 
     def delayed_events
--- a/lib/dynflow/testing/dummy_executor.rb
+++ b/lib/dynflow/testing/dummy_executor.rb
@@ -17,6 +17,18 @@ module Dynflow
         @events_to_process << [director_event.execution_plan_id, director_event.step_id, director_event.event, director_event.result]
       end
 
+      def plan_events(delayed_events)
+        delayed_events.each do |event|
+          world.plan_event(event.execution_plan_id, event.step_id, event.event, event.time)
+        end
+      end
+
+      def execute(action, event = nil)
+        action.execute event
+        plan_events(action.delayed_events.dup)
+        action.delayed_events.clear
+      end
+
       # returns true if some event was processed.
       def progress
         events = @events_to_process.dup
@@ -26,7 +38,7 @@ module Dynflow
           if event && world.action.state != :suspended
             return false
           end
-          world.action.execute event
+          execute(world.action, event)
         end
       end
 
--- a/test/action_test.rb
+++ b/test/action_test.rb
@@ -570,9 +570,9 @@ module Dynflow
               total = 2
               FailureSimulator.fail_in_child_plan = true
               triggered_plan = world.trigger(PollingParentAction, count: total)
-              polling_plan = world.persistence.load_execution_plan(triggered_plan.id)
 
-              wait_for do # Waiting for the sub plans to be spawned
+              polling_plan = nil
+              wait_for('the subplans to be spawned') do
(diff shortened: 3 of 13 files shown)


bin/dynflow_sidekiq.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
```
adamruzicka: This file is not really relevant

lib/dynflow/action/with_sub_plans.rb
```diff
@@ -170,10 +170,10 @@ def sub_plans_count(filter = {})
     end
 
     def notify_on_finish(plans)
-      suspend do |suspended_action|
+      suspend do
         plans.each do |plan|
           plan.finished.on_resolution! do |success, value|
-            suspended_action << SubPlanFinished[plan.id, success && (value.result == :success)]
+            plan_event(SubPlanFinished[plan.id, success && (value.result == :success)])
```
adamruzicka: In this case we should probably keep it as it is, as the new way wouldn't work once the action gets suspended.

ezr-ondrej: Yeah right, we would need to propagate the information about parent action to through the execution plan and ping from there :(

adamruzicka:
Test failures seem to be related

test/test_helper.rb
```diff
@@ -177,7 +177,7 @@ def wait_for(waiting_message = 'something to happen')
       return ret if ret
       sleep 0.3
     end
-    raise "waiting for #{waiting_message} was not successful"
+    assert false, "waiting for #{waiting_message} was not successful"
```
ezr-ondrej: I've changed this to assert - it should fail not error out I believe.

test/action_test.rb
```diff
@@ -570,9 +570,9 @@ def batch(from, size)
               total = 2
               FailureSimulator.fail_in_child_plan = true
               triggered_plan = world.trigger(PollingParentAction, count: total)
-              polling_plan = world.persistence.load_execution_plan(triggered_plan.id)
 
-              wait_for do # Waiting for the sub plans to be spawned
+              polling_plan = nil
```
ezr-ondrej: ambigous loading removed

lib/dynflow/testing/dummy_executor.rb
```diff
@@ -17,6 +17,20 @@ def delayed_event(director_event)
...
 
+      def plan_events(delayed_events)
+        delayed_events.each do |event|
+          time = event.time
+          time = world.clock.current_time + time if time.is_a?(Numeric)
+          world.plan_event(event.execution_plan_id, event.step_id, event.event, time)
+        end
+      end
+
+      def execute(action, event = nil)
```
ezr-ondrej: New method called from `Testing#run_action` and `progress` as both need to process action's events

lib/dynflow/action.rb
```diff
@@ -89,6 +89,13 @@ def to_s_humanized
       end
     end
 
+    DelayedEvent = Algebrick.type do
```
ezr-ondrej: New `DelayedEvent` class, for not abusing `Director::Event` and to validate the time properly.

lib/dynflow/action.rb
```diff
@@ -89,6 +89,13 @@ def to_s_humanized
...
+      fields! execution_plan_id: String,
+              step_id:           Integer,
+              event:             Object,
+              time:              type { variants Numeric, NilClass }
```
ezr-ondrej: :warning: On the action class I've removed option to pass `Time` only a number, do you think it's ok?

lib/dynflow/world.rb
```diff
@@ -228,6 +228,7 @@ def event(execution_plan_id, step_id, event, done = Concurrent::Promises.resolva
     end
 
     def plan_event(execution_plan_id, step_id, event, time, accepted = Concurrent::Promises.resolvable_future)
+      time = Time.now + time if time.is_a?(Numeric)
```
ezr-ondrej: Numerical `time` param is processed on orchestrator side now.
:warning: this could delay the events for the communication error! Is that acceptable?

ezr-ondrej:
And we are :green_heart:

lib/dynflow/world.rb
adamruzicka: It shouldn't be an issue when things are working smoothly, but I'd be worried what would happen once the load got high enough. What was the reason for this change?

ezr-ondrej: The reason is just tests - with testing clock, we can't have real time, but probably we can think of better solution :thinking:

adamruzicka: I'd like to keep the option to send times, if possible. Is it doable in the time we have left?

ezr-ondrej: Ok, I did it through the clock. It now provides `current_time` even in production. Providing wrapper arround Time.now

lib/dynflow/action.rb
```diff
@@ -326,9 +333,8 @@ def queue
     # Plan an +event+ to be send to the action defined by +action+, what defaults to be self.
     # if +time+ is not passed, event is sent as soon as possible.
     def plan_event(event, time = nil, execution_plan_id: self.execution_plan_id, step_id: self.run_step_id)
-      Type! time, Time, Numeric, NilClass
-      time  = Time.now + time if time.is_a? Numeric
-      delayed_events << Dispatcher::Event[execution_plan_id, step_id, event, time]
+      time = @world.clock.current_time + time if time.is_a?(Numeric)
```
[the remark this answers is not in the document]
ezr-ondrej: Yeah, I did, we pass that down to the `DelayedEvent` and we are validating the type of the field there, right? :slightly_smiling_face:

adamruzicka:
Thank you @ezr-ondrej !

PR #348, opened 2019-11-26
Tests for failover HA

adamruzicka:


test/redis_locking_test.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      let(:world) { OpenStruct.new(:id => '12345') }
+      let(:orchestrator) { Orchestrator.new(world, Logger.new) }
+      let(:world2) { OpenStruct.new(:id => '67890') }
+      let(:orchestrator2) { Orchestrator.new(world2, Logger.new) }
+
+      it 'acquires the lock when it is not taken' do
+        orchestrator.wait_for_orchestrator_lock
+        logs = orchestrator.logger.logs
+        _(logs).must_equal [[:info, 'Acquired orchestrator lock, entering active mode.']]
```
ezr-ondrej: Do you think it would be possible to test the value in the redis directly (as we have the connection already, we could use it to assert the orchestrator behaviour, right?

adamruzicka: Good point, added.

adamruzicka:
I'm not sure what else we could test apart from the lock value without making the test too complex

PR #349, opened 2019-12-03
Reimplement various "what-to-do-next" components using dependency graph

adamruzicka:
The idea is that the new representation should allow us to store this kind of data in a db (be it redis or a full blown sql db).


adamruzicka:
I always had trouble wrapping my head around how dynflow decides what to do next and how it processes the flows. I took some time to throw most of it out and use a directed acyclic graph as an intermediate representation and then plucking leaf nodes out of it. Currently it still uses flows when creating an execution plan, but in the future I'd like to remove them and just use the graph everywhere.

Although readability is the primary concern here, this should also address one situation where the old mechanism executed steps in sequence even when it didn't have to.

@jlsherrill I need an impartial set of eyes to judge if my efforts here actually helped readability and ease of understanding. Could you please take a look?

Just don't look too hard at [graph->flow conversion](https://github.com/Dynflow/dynflow/pull/349/files#diff-d9d66869d60c44274bfacc5de7294bdf9197ff668a7462320f37265ed20a20ddR105), which is terrible, but would go away if we used the graph everywhere

PR #350, opened 2020-02-28
update concurrent-ruby-edge dependency

mmoll:


mmoll:
Maybe this could also be `>= 0.4.1`?

adamruzicka:
Let's keep it at `~> 0.6.0`

adamruzicka:
Thank you @mmoll!

PR #351, opened 2020-03-23
Cleanup of undeliverable envelopes

adamruzicka:


ezr-ondrej:
Tests are failing, otherwise the idea makes sense :) Should I go ahead and test even the tests are failing?

lib/dynflow/connectors/direct.rb
```diff
@@ -34,10 +39,6 @@ def handle_envelope(envelope)
           end
         end
 
-        def prune_undeliverable_envelopes(_world)
```
adamruzicka: This is why I don't like having several classes in a single file

lib/dynflow/persistence_adapters/sequel.rb
```diff
@@ -198,6 +198,16 @@ def push_envelope(envelope)
...
       end
 
+      def prune_envelopes(receiver_ids)
+        connector_feature!
+        table(:envelope).where(receiver_id: receiver_ids).delete
+      end
+
+      def prune_undeliverable_envelopes
+        connector_feature!
+        table(:envelope).where(receiver_id: table(:coordinator_record).select(:id)).invert.delete
```
ezr-ondrej, line 208: This means we don't expect two coordinators running on top of the same database, right?
Could it be a problem in the future?

adamruzicka: Why would that be a problem?

ezr-ondrej: If we decide we want multiple coordinators for some reason.

adamruzicka: We already have multiple coordinators, in production deployments of Foreman you get a couple of them.

In order to be considered valid, it has to have a record in `dynflow_coordinator_records` table. This should affect only envelopes which don't have a receiver in that table. I'm still not following what the issue might be

ezr-ondrej: Oh, right, I've not read correctly the `invert`.

lib/dynflow/connectors/direct.rb
```diff
@@ -3,6 +3,11 @@ module Dynflow
   module Connectors
     class Direct < Abstract
 
+      def prune_undeliverable_envelopes(_world)
```
ezr-ondrej, line 6: Could we move this one after the `Core` class definition with the other class methods?
Not sure if it's the correct place, but at least to be consistent :)

ezr-ondrej:
Thanks @adamruzicka!

PR #357, opened 2020-06-23
Use correct method name for resolving events

adamruzicka:
Concurrent-ruby-1.1.3 changed a lot of method names around futures and
events. Apparently we missed one of those when adjusting the code to be
compatible with this version.

### DIFF
--- a/lib/dynflow/world.rb
+++ b/lib/dynflow/world.rb
@@ -341,7 +341,7 @@ module Dynflow
         @terminating = Concurrent::Promises.future do
           termination_future.wait(termination_timeout)
         end.on_resolution do
-          @terminated.complete
+          @terminated.resolve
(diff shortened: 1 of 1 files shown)


adamruzicka:
@hao-yu hi, could you take a look?

hao-yu:
@adamruzicka Thanks for the patch. Do you still need to call "terminated.resolve" again in the on_resolution when you already call it in https://github.com/Dynflow/dynflow/pull/357/files#diff-35ca92bf5b10414a154a2844535ab5e9R335 ?

adamruzicka:
if an exception is raised anywhere in the block around L335, then `@terminated` would not be resolved. And considering calling `resolve` on an already resolved event is a no-op, I'd say it is ok to keep it there

hao-yu:
@adamruzicka looks good to me.

PR #358, opened 2020-06-25
Store flow data in a more compact manner

adamruzicka:
This new format is roughly 10 times more efficient than what we had previously.

### Old format
```json
{
  "class": "Dynflow::Flows::Concurrence",
  "flows": [
    {
      "class": "Dynflow::Flows::Atom",
      "step_id": 3
    },
    {
      "class": "Dynflow::Flows::Sequence",
      "flows": [
        {
          "class": "Dynflow::Flows::Atom",
          "step_id": 4
        },
        {
          "class": "Dynflow::Flows::Atom",
          "step_id": 7
        }
      ]
    },
    {
      "class": "Dynflow::Flows::Atom",
      "step_id": 10
    }
  ]
}
```

### New format
```json
["C",  3,
       ["S", 4, 7],
       10]
```

Note: both examples were reformatted for readability, in the db they are just `MultiJson.dump`ed

### DIFF
--- a/lib/dynflow/flows/abstract.rb
+++ b/lib/dynflow/flows/abstract.rb
@@ -41,14 +41,7 @@ module Dynflow
           Flows::Atom.new(hash)
         else
           kind, *subflows = hash
-          klass = case kind
-                  when "S"
-                    Sequence
-                  when "C"
-                    Concurrence
-                  else
-                    raise("Unknown composed flow type")
-                  end
+          klass = AbstractComposed::FLOW_SERIALIZATION_MAP[kind] || raise("Unknown composed flow type")
           klass.new(subflows.map { |subflow| self.new_from_hash(subflow) })
         end
       end
--- a/lib/dynflow/flows/abstract.rb
+++ b/lib/dynflow/flows/abstract.rb
@@ -41,8 +41,7 @@ module Dynflow
           Flows::Atom.new(hash)
         else
           kind, *subflows = hash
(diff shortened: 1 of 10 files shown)


adamruzicka:
Better now?

ezr-ondrej:
I'd maybe go with a global registery as if I want to register new type, I need to see what letters are available. This way, it won't even detect the clash of two classes, rigth? It would just roll with the last registered class :thinking:

lib/dynflow/flows/atom.rb
```diff
@@ -6,7 +6,7 @@ class Atom < Abstract
       attr_reader :step_id
 
       def to_hash
```
ezr-ondrej, line 8: The naming is quite weird now... :thinking:
Should we introduce new method? e.g. `dump` or `encode`?

lib/dynflow/flows/abstract.rb
```diff
@@ -32,6 +32,18 @@ def all_step_ids
       def flatten!
         raise NotImplementedError
       end
+
+      def self.new_from_hash(hash)
```
ezr-ondrej, line 36: are we able to rename this method as `load` or are we bound to this name by call from another structures?

adamruzicka: Switched to `encode` and `decode`

ezr-ondrej:
Thanks @adamruzicka ! I believe this is a great feature!

PR #359, opened 2020-06-30
Store free-form data as msgpack

adamruzicka:


adamruzicka:
@ezr-ondrej This is now ready. I'd like to get it in shortly after Foreman 2.6 branching so it is ready for prime time in 3.0.

ezr-ondrej:
> @ezr-ondrej This is now ready. I'd like to get it in shortly after Foreman 2.6 branching so it is ready for prime time in 3.0.

UGH? 2.6 = 3.0 :thinking:

ezr-ondrej:
I'll take if for a spin :)

adamruzicka:
> 2.6 = 3.0

Oh. Nevermind the specifics, let's get it in sometime during this year

ezr-ondrej:
Could this be related: https://pastebin.com/54G9kQr6 ?

adamruzicka:
On my machine with a database with 8337 execution plans, 143659 actions and 244454 steps the migration took 261.995207 seconds. Not all that great considering the db wasn't that big

adamruzicka:
On my machine with a database with 58385 execution plans, 373500 actions and 689464 the migration took 728.806448 seconds which is not that bad.

ezr-ondrej:
Not, great, but doable I think. The gain is worth the waiting, if you have a huge db, you'll save a lot of space for the time you've waited :shrug:

ezr-ondrej (approved):
Times seems reasonable to me :+1:

adamruzicka:
> Could this be related: https://pastebin.com/54G9kQr6 ?
> EDIT: Might have been me first running some jobs with the patch applied and only then running migration thinking

So this was a real issue after all https://github.com/Dynflow/dynflow/pull/397

PR #361, opened 2020-08-04
Fix undefined method 'result' error on termination

adamruzicka:


ezr-ondrej (changes requested):
Hmmm EventWorkItem has an event field and event field has result and result is a `Concurrent::Promises::ResolvableFuture`. It would seem to me more like the event field is nil in some cases, but this doesn't seem like a proper solution.

ezr-ondrej (approved):
Thanks @adamruzicka !

PR #363, opened 2020-08-07
Do not try to change plan's state if it is in it already

adamruzicka:
This should prevent errors like invalid transition stopped >> stopped during
world invalidation.


adamruzicka:
The tests around invalidation live mostly in `test/abnormal_states_recovery_test.rb`

ezr-ondrej (approved):
Thanks @adamruzicka ! :+1:

PR #364, opened 2020-08-11
Reject pings properly if receiver is executor and cache is stale

adamruzicka:


ezr-ondrej (approved):
Thanks @adamruzicka ! :+1:

PR #365, opened 2020-09-03
Groundwork for yaml export

adamruzicka:


adamruzicka:
Something tells me this is not going to happen, closing.

PR #366, opened 2020-10-12
Dispatch delayed plan planning through connector 

adamruzicka:
Before this patch, planning of delayed plans was done by the delayed executor
actor, which runs in the orchestrator process. This had several drawbacks. The
polling interval used by the delayed executor could increase when it was
performing planning, it could have required additional ActiveRecord connection
and it required the entire application environment to be loaded.

With this change, the actual planning is dispatched through connector to the
workers. To prevent planning a plan twice, a flag is set on the delayed plan
record to indicate it is already being handled. Deletion of delayed plan record
is moved to the worker performing the actual planning.

This also allows us to "plan elsewhere", kick off a task but let another process
perform the plan phase without blocking the client with planning.

Addresses #371

### DIFF
--- a/lib/dynflow/delayed_executors/abstract_core.rb
+++ b/lib/dynflow/delayed_executors/abstract_core.rb
@@ -46,24 +46,21 @@ module Dynflow
 
       def process(delayed_plans, check_time)
         processed_plan_uuids = []
+        dispatched_plan_uuids = []
+        planning_locks = world.coordinator.find_records(class: Coordinator::PlanningLock.name)
         delayed_plans.each do |plan|
-          next if plan.frozen
+          next if plan.frozen || locked_for_planning?(planning_locks, plan)
           fix_plan_state(plan)
           with_error_handling do
             if plan.execution_plan.state != :scheduled
               # in case the previous process was terminated after running the plan, but before deleting the delayed plan record.
               @logger.info("Execution plan #{plan.execution_plan_uuid} is expected to be in 'scheduled' state, was '#{plan.execution_plan.state}', skipping")
-            elsif !plan.start_before.nil? && plan.start_before < check_time
-              @logger.debug "Failing plan #{plan.execution_plan_uuid}"
-              plan.timeout
+              processed_plan_uuids << plan.execution_plan_uuid
             else
               @logger.debug "Executing plan #{plan.execution_plan_uuid}"
-              Executors.run_user_code do
-                plan.plan
-                plan.execute
-              end
+              world.plan_request(plan.execution_plan_uuid)
+              dispatched_plan_uuids << plan.execution_plan_uuid
             end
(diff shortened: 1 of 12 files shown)


lib/dynflow/delayed_executors/abstract_core.rb
```diff
@@ -46,32 +46,30 @@ def with_error_handling(error_retval = nil, &block)
...
           end
         end
+        world.persistence.mark_delayed_plans_as_planning(:execution_plan_uuid => dispatched_plan_uuids) unless dispatched_plan_uuids.empty?
         world.persistence.delete_delayed_plans(:execution_plan_uuid => processed_plan_uuids) unless processed_plan_uuids.empty?
       end
 
       private
 
       # handle the case, where the process was termintated while planning was in progress before
+      # TODO: Doing execution plan updates in orchestrator is bad
```
ezr-ondrej, line 72: :laughing:

adamruzicka:
Oh, this needs a rebase now, will do shortly

ezr-ondrej (commented):
Do we strictly need the `planning`  flag? wouldn't entering the Planning stage be enough? and just checking we're not Planning already before the worker really starts on that?

Unless of course we are entering the Planning on creating the execution plan.

adamruzicka:
Currently when polling for delayed plans we only look into `dynflow_delayed_plans` table. If we wanted to rely on the execution plan's state, we'd have to start joining the tables. It would also mean that we might submit a planning request multiple times if the previous one wasn't processed in time.

ezr-ondrej:
> Currently when polling for delayed plans we only look into `dynflow_delayed_plans` table. If we wanted to rely on the execution plan's state, we'd have to start joining the tables. It would also mean that we might submit a planning request multiple times if the previous one wasn't processed in time.

Yeah, I wouldn't see problem in sending the request to multiple workers if we can make sure it will be processed just once. The same with run steps. It's just a thought obviously... flag is not that bad, but I feel it is a bit redundant (especially given we want to plan all the plans in the worker, so IIUC the flag should really just note "has it been sent out for planning yet?"

adamruzicka:
Dropped the flag, noticed we don't create planning locks when planning delayed plans so I added that as well.

ezr-ondrej (approved):
I'm ok with this and it is a step forward.
Holding on merging as I'm not sure how we are on releasing, but getting this in with 2.6 foreman would be nice.

adamruzicka:
> Holding on merging as I'm not sure how we are on releasing

We're currently clean. 1.4.8 for <=2.5 and 1.5.0 for 2.6+ are both out. Feel free to merge anytime you feel like it

adamruzicka:
Alright, merging

adamruzicka:
Thank you @ezr-ondrej for the review

PR #377, opened 2021-01-04
Try running tests with ruby 3.0.0

adamruzicka:


adamruzicka:
Looks like it mostly works, apart from the timing issues (`Expected |2021-01-04 10:08:20 UTC - 2021-01-04 10:08:20 UTC| (0.944137897) to be <= 0.5.`) we've been encountering since moving to GH actions.

PR #378, opened 2021-01-22
Implement message bus based connector

adamruzicka:
Historically Dynflow used database for communication across processes. The
sending side would write a record into dynflow_envelopes table, receiving side
would poll this table and if it found a record addressed to it, it would receive
it and remove it from the db. Both polling and insert-read-delete cycles were
putting unnecessary load on the db. Using a message bus for this should be the
way to go.

Here I chose MQTT because it is lightweight and I worked with it recently and
still have it in fresh memory. However, this could be adjusted to almost
anything with minimal changes.

#373

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,3 +42,7 @@ end
 group :telemetry do
   gem 'statsd-instrument'
 end
+
+group :mqtt do
+  gem 'mqtt'
+end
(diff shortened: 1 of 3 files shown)


adamruzicka:
I also considered using Redis since we kind of have it already, but there we would either duplicate connections per-process if running with Sidekiq or have slightly different behavior when running with and without Sidekiq. Another gotcha is pub-sub in redis is not scoped to databases. It might not be an issue, but definitely isn't a plus

PR #379, opened 2021-02-05
fix: invalid constant exception

majioa:
- [x] ! invalid constant exception Railties due to invalid context search


adamruzicka:
Hi, thanks for the patch, in general +1. How would I go about reproducing the issue?

adamruzicka:
Thank you @majioa !

PR #380, opened 2021-02-08
Zeitwerk

adamruzicka:


adamruzicka:
Tests are as spotty as always, but this should be good to go.

ofedoren:
@adamruzicka, one of the tests failed, is this expected?

adamruzicka:
Some of the tests test that "something happens within given time" and the time set there should be more than enough if run on a machine made in this century. Sadly sometimes it feels like github actions run on a toaster from 1980s so some of the tests sometimes fail.

On a side note, it would be worth testing this on a production setup with foreman and katello. It seems to work fine for me locally, but I can't say I understand zeitwerk enough to be able to rule out any hidden dragons.

lib/dynflow.rb
```diff
@@ -13,6 +13,22 @@
   logger.add level, message, progname, &block
 end
 
+module Dynflow
+end
+
+require 'zeitwerk'
+loader = Zeitwerk::Loader.for_gem
```
ofedoren, line 20: This looks odd based on the docs: https://github.com/fxn/zeitwerk/tree/main#for_gem, but it somehow works...

ofedoren (commented):
Thanks, @adamruzicka! I've checked it on the production setup and it seems to load just fine (with zeitwerk 2.5.3, it didn't work with zeitwerk 2.2.2). Wasn't able to test all the things though due to some other broken things not related to this PR...

PR #382, opened 2021-03-15
Fix uninitialized constant Dynflow::Rails::Railtie

ehelms:


ehelms:
I found this doing some Ruby 2.7 testing. This does not block anything that I can find, but seems like it can appear under certain other failures and cause false positives.

adamruzicka:
Thank you for the patch, but I'll close it in favor of https://github.com/Dynflow/dynflow/pull/379 which is identical, but was opened first.

PR #384, opened 2021-03-30
Fix frozen literal in inspect

lzap:
Strings are frozen for that file, therefore appending will not work and result in exception when inspect is called.

### DIFF
--- a/lib/dynflow/execution_history.rb
+++ b/lib/dynflow/execution_history.rb
@@ -12,7 +12,7 @@ module Dynflow
 
     module Event
       def inspect
-        "#{Time.at(time).utc}: #{name}".tap { |s| s << " @ #{world_id}" if world_id }
+        [ "#{Time.at(time).utc}: #{name}", world_id ].compact.join(' @ ')
       end
     end
 


lib/dynflow/execution_history.rb
```diff
@@ -12,7 +12,11 @@ class ExecutionHistory
 
     module Event
       def inspect
-        "#{Time.at(time).utc}: #{name}".tap { |s| s << " @ #{world_id}" if world_id }
+        if world_id
+          "#{Time.at(time).utc}: #{name} @ #{world_id}"
+        else
+          "#{Time.at(time).utc}: #{name}"
+        end
```
adamruzicka, line 19: Could we do something like this to reduce repetition?
```suggestion
        [ "#{Time.at(time).utc}: #{name}", world_id ].compact.join(' @ ')
```

lzap: I prefer faster code, but this is indeed more of a debugging part so why not.

lzap:
Rebased

adamruzicka:
Rubocop is unhappy
```
lib/dynflow/execution_history.rb:15:10: C: Space inside square brackets detected.
        [ "#{Time.at(time).utc}: #{name}", world_id ].compact.join(' @ ')
         ^
lib/dynflow/execution_history.rb:15:52: C: Space inside square brackets detected.
        [ "#{Time.at(time).utc}: #{name}", world_id ].compact.join(' @ ')
                                                   ^
```

adamruzicka:
Fixed rubocop and merged by hand as 6c6bca2ade7b11356e6bb108257e48a49e23abdd , thank you @lzap!

lzap:
I am getting there, I am getting there... I think I need to unsubscribe some repos.

Thanks.

PR #387, opened 2021-04-30
Add support for chunked outputs

adamruzicka:
The benchmark reads a list of words and then stores first `$count` words as the action's output, grouping them into groups of `$group` words.

```
bundle exec ruby examples/chunked_output_benchmark.rb                                                                                                                                                                     
World 9c3f42c8-e828-471e-8ded-a9fa9f9378e5 started...
                           user       system     total    real
regular    1000 by    100  0.052024   0.006064   0.058088 (  0.049565)
chunked    1000 by    100  0.041618   0.007773   0.049391 (  0.041637)
regular  10_000 by    100  1.085945   0.070478   1.156423 (  1.097117)
chunked  10_000 by    100  0.481077   0.030642   0.511719 (  0.458789)
regular  10_000 by   1000  0.117079   0.001061   0.118140 (  0.113168)
chunked  10_000 by   1000  0.045330   0.004583   0.049913 (  0.042220)
regular 100_000 by    100 70.071629   0.730572  70.802201 ( 70.252973)
chunked 100_000 by    100  8.764937   0.599450   9.364387 (  8.613024)
regular 100_000 by   1000  6.425341   0.028132   6.453473 (  6.389819)
chunked 100_000 by   1000  0.470149   0.021370   0.491519 (  0.423788)
regular 100_000 by 10_000  0.760500   0.001779   0.762279 (  0.754986)
chunked 100_000 by 10_000  0.067514   0.005190   0.072704 (  0.063342)
```

Looking at the `chunked` lines we can see the time grows almost linearly with `$count/$group`, but for the regular output it grows much faster.

### DIFF
--- a/web/views/flow_step.erb
+++ b/web/views/flow_step.erb
@@ -31,6 +31,7 @@
   <% end %>
   <%= show_action_data("Input:", action.input) %>
   <%= show_action_data("Output:", action.output) %>
+  <%= show_action_data("Chunked output:", action.stored_output_chunks) %>
   <% if step.error %>
     <p>
       <b>Error:</b>
--- a/examples/chunked_output_benchmark.rb
+++ b/examples/chunked_output_benchmark.rb
@@ -1,0 +1,77 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative 'example_helper'
+require 'benchmark'
+
+WORDS = File.readlines('/usr/share/dict/words').map(&:chomp).freeze
+COUNT = WORDS.count
+
+module Common
+  def main_loop
+    if output[:current] < input[:limit]
+      consumed = yield
+      output[:current] += consumed
+      plan_event(nil)
+      suspend
+    end
+  end
+
+  def batch
+    WORDS.drop(output[:current]).take(input[:chunk])
+  end
+end
+
+class Regular < ::Dynflow::Action
+  include Common
+
+  def run(event = nil)
+    output[:current] ||= 0
+    output[:words] ||= []
+
+    main_loop do
+      words = batch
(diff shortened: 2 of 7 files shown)


web/views/flow_step.erb
```diff
@@ -31,6 +31,7 @@
   <% end %>
   <%= show_action_data("Input:", action.input) %>
   <%= show_action_data("Output:", action.output) %>
+  <%= show_action_data("Chunked output:", action.stored_output_chunks) %>
```
ezr-ondrej, line 34: Should we display this to the user, shouldn't we join the chunks when showing?

ezr-ondrej:
I like this funcionality, but I'm unsure about the user interface for this.

adamruzicka:
> I'm unsure about the user interface for this.

I admit it is not the best. I'm open to ideas

web/views/flow_step.erb
adamruzicka: Can we do that? I mean we provide an option to store data in an append only fashion, but we don't really care what the data is. Some of it may not be join-able

ezr-ondrej: Got it :+1:

ezr-ondrej:
> I'm open to ideas

Not sure without proper design. I'm not clear on how it is supposed to be used, so we can get this in and see after we have users for it.

adamruzicka:
Have you seen https://github.com/theforeman/foreman-tasks/pull/633/files and https://github.com/theforeman/smart_proxy_dynflow/pull/80/files ? They are rather bare bones, but that's all that is needed.

ezr-ondrej (approved):
Thanks @adamruzicka, let get this in

PR #389, opened 2021-05-04
Do not collect full stacktraces by default

### LINKED ISSUE #388: Dynflow collects stack trace across action suspends
If an action suspends itself and on resume raises an exception, the printed stacktrace does not contain only trace from the resume to the place where the exception was raised, but it also includes the part leading to the suspend. Additionally, the traces accumulate over any number of suspends and resumes. If an action suspends, resumes, suspends and then raises an exception, the trace goes from start to suspend, from resume to suspend and finally from resume to raise. While this might be useful when developing dynflow internals, its usefulness in production deployments is debatable at best.

This is especially harmful with long running polling actions, which can accumulate huge amount of stack traces while they're active. And even if the stacktrace isn't printed, it is still collected leading to memory bloat.

For reproducer see [gist](https://gist.github.com/adamruzicka/950f005f0a817fd08dfaed635912b3b7)

The formula for number of lines being emitted is `lines = 53 * $SUSPEND_COUNT + 76`, formula for output size is `bytes = 4415 * SUSPEND_COUNT + 6470`.

| Number of suspends before raise | Stacktrace length [lines] | Stacktrace size [bytes] |
| -------------: | -----:| -----: |
| 1       | 129 | 10885 |
| 5 | 341 | 28545 |
| 10 | 606 | 50620 |
| 50 | 2726 | 227220 |
| 100 | 5376 | 447970 |
| 500 | 26576 | 2213970 |
| 1000 | 53076 | 4421470 |

I'm not sure how much of an issue is this in sidekiq-based deployments where events actually cross process borders.

BZ https://bugzilla.redhat.com/show_bug.cgi?id=1877917

adamruzicka:
If desired, their collection can be enabled by setting `DYNFLOW_FULL_BACKTRACE` environment variable. Additionally, the length of the output can be limited by setting `DYNFLOW_BACKTRACE_LIMIT`.

Fixes #388

### DIFF
--- a/lib/dynflow/actor.rb
+++ b/lib/dynflow/actor.rb
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 module Dynflow
 
+  FULL_BACKTRACE = %w[1 y yes].include?((ENV['DYNFLOW_FULL_BACKTRACE'] || '').downcase)
+  BACKTRACE_LIMIT = begin
+                      limit = ENV['DYNFLOW_BACKTRACE_LIMIT'].to_i
+                      limit.zero? ? nil : limit
+                    end
+
   module MethodicActor
     def on_message(message)
       method, *args = message
@@ -44,7 +50,11 @@ module Dynflow
       include LogWithFullBacktrace
 
       def on_envelope(envelope)
-        Actor::BacktraceCollector.with_backtrace(envelope.origin_backtrace) { super }
+        if FULL_BACKTRACE
+          Actor::BacktraceCollector.with_backtrace(envelope.origin_backtrace) { super }
+        else
+          super
+        end
       end
     end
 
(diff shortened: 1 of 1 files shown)


ezr-ondrej (commented):
Generally I like it, I usually have hard time to find the real exception as there is a lot of internal dynflow stuff in the backtrace.
Would it also help to improve the filter?
Would it be possible to make this settable during the runtime? or is this mostly for workers, so they should just get restarted if change is needed?

adamruzicka:
> Would it also help to improve the filter?

I'm open to ideas, what would you like to see filtered out? Or make the filter itself configurable as an env var?

> Would it be possible to make this settable during the runtime?

I could imagine toggling it on SIGUSR1 or something, although it could lead to incomplete stacktraces for the thing which are just being executed.

> or is this mostly for workers, so they should just get restarted if change is needed?

In production this is almost exclusively for workers. In development where people run an executor inside the rails process it could be more useful, otoh restarting the server in development is not a big deal imo (and most time even necessary).

ezr-ondrej (approved):
I'm fine with this, I consider this more of a workaround then a fix.
I've got no clear idea what should be filtered out or how to fix the core issue (maybe filter out repetitions somehow? :shrug:

PR #391, opened 2021-09-06
Add run_step_id to DummyAction

ezr-ondrej:


adamruzicka:
In general +1, let's see what the tests think

ezr-ondrej:
> In general +1, let's see what the tests think

In theory tests should not touch the method as they do not need it.

ezr-ondrej:
Just mysql, which is broken long term :)

adamruzicka:
Thank you @ezr-ondrej !

PR #393, opened 2021-09-17
Run hooks when execution plan is first created

adamruzicka:
TODO:
- [x] tests

### DIFF
--- a/lib/dynflow/execution_plan/hooks.rb
+++ b/lib/dynflow/execution_plan/hooks.rb
@@ -21,7 +21,7 @@ module Dynflow
         # @param class_name [Class] class of the hook to be run
         # @param on [Symbol, Array<Symbol>] when should the hook be run, one of {HOOK_KINDS}
         # @return [void]
-        def use(class_name, on: HOOK_KINDS)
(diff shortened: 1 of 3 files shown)


lib/dynflow/execution_plan/hooks.rb
```diff
@@ -21,7 +21,7 @@ def initialize(hooks = {})
         # @param class_name [Class] class of the hook to be run
         # @param on [Symbol, Array<Symbol>] when should the hook be run, one of {HOOK_KINDS}
         # @return [void]
-        def use(class_name, on: HOOK_KINDS)
+        def use(class_name, on: ExecutionPlan.states)
```
ezr-ondrej, line 24: Why we did add `:success`, `:failure` before and now we do not need to? :thinking:

adamruzicka: I wanted to make this run the hook on all state transitions. What I overlooked was that `:success` is a shorthand for `stopped-success` state-result pair, which meant that the hook was being run twice if the plan stopped successfully.

ezr-ondrej (approved):
Thanks @adamruzicka !

PR #394, opened 2021-09-20
Fix tests on MySQL

adamruzicka:


test/future_execution_test.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require_relative 'test_helper'
+require 'multi_json'
```
ezr-ondrej, line 3: This doesn't seem necessary :thinking:

ezr-ondrej (approved):
Apart of the uneccessary require, I like this :+1:

test/future_execution_test.rb
adamruzicka: I needed to do this when running tests only in this file, will drop.

ezr-ondrej:
Thanks @adamruzicka !

PR #397, opened 2021-09-23
Fix undefined method `inject' for nil:NilClass when loading plans

adamruzicka:


ezr-ondrej:
Shouldn't just the last clause be else instead of empty `when`?
What I've got no idea what should do, but I'd imagine else there and then the first case not needed.

adamruzicka:
It would be correct the split  the last bit off into a new migration. Do we want to be that correct, given 1.6.0 didn't have that much time to spread?

ezr-ondrej:
No we don't lets cut a new release and yank `1.6.0`

PR #398, opened 2021-09-24
Do not update condition keys

ezr-ondrej:
Condition keys, usually e.g. PKs are updated on every update, that might cause issues.


adamruzicka:
IIRC it turned out we don't really need this, do we?

PR #399, opened 2021-09-24
Fix odd behavior on sqlite 3.7.17

adamruzicka:
Older sqlite raise foreign key constraint errors even though the data is valid.
The same code works with newer sqlite so I guess it is a bug in sqlite which was
fixed in the meantime.

### DIFF
--- a/lib/dynflow/persistence_adapters/sequel_migrations/023_sqlite_workarounds.rb
+++ b/lib/dynflow/persistence_adapters/sequel_migrations/023_sqlite_workarounds.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+tables = [:dynflow_actions, :dynflow_delayed_plans, :dynflow_steps, :dynflow_output_chunks]
+Sequel.migration do
+  up do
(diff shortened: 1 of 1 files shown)


adamruzicka:
Tests on EL7 running here https://github.com/theforeman/smart_proxy_dynflow/pull/98

lib/dynflow/persistence_adapters/sequel_migrations/023_sqlite_workarounds.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+Sequel.migration do
+  up do
+    if database_type == :sqlite && SQLite3::SQLITE_VERSION == "3.7.17"
```
ezr-ondrej, line 4: should we have that specific version? wouldn't it be better to have older than?
Or just drop it on SQLite always? There is not much value in having constraints that works part of the time I guess.

ezr-ondrej (approved):
Feel free to merge if you think the versioning is fine :)

adamruzicka:
Dropped the version check, tests in smart_proxy_dynflow passed, waiting for tests here to turn green

adamruzicka:
Tests are finally green, let's get this in.

PR #401, opened 2022-01-07
Add workflow to build go tooling on tag push

adamruzicka:


adamruzicka:
It generates releases like this one https://github.com/adamruzicka/dynflow/releases/tag/v1.6.3

.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
...
+        with:
+          tagRegex: "v(.*)"  # Optional. Returns specified group text as tag name. Full tag string is returned if regex is not defined.
+          tagRegexGroup: 1 # Optional. Default is 1.
+      - name: Checkout the repository
+        uses: actions/checkout@v2
+      - name: Generate build files
+        run: |
+          mkdir -p dist
+          cd extras/expand
+          go build -o ../../dist/dynflow-expand-${{ steps.tagName.outputs.tag }}-x86_64
```
ezr-ondrej, line 27: I'd rather see VERSION used here as well, or not used in the other task as well, but its nit :)

ezr-ondrej (approved):
The GH action looks good and I like removing code :thinking:

.github/workflows/release.yml
adamruzicka: Done

adamruzicka:
I addressed the last comment and squashed the commits a bit

ezr-ondrej (approved):
Thanks @adamruzicka ! :+1:

PR #402, opened 2022-01-10
Use JSON directly in json to msgpack migration

adamruzicka:
MultiJson provides a common interface over several JSON parsers. Sadly some of
the parsers have subtle differences.

```ruby
require 'json'
require 'oj'

example = '{"key": 1637704901.3995261}'
Oj.load(example)['key'].class # => BigDecimal
JSON.parse(example)['key'].class # => Float
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,7 +35,7 @@ end
 
 group :rails do
   gem 'daemons'
-  gem 'rails', '>= 4.2.9'
+  gem 'rails', '>= 4.2.9', '< 7'
   gem 'logging'
 end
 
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
From IRC `#theforeman`:
```
<aruzicka> KeeperoftheKeys: hi, I have a patch ready for the issue you asked about yesterday
<KeeperoftheKeys> cool! would you like me to test it?
<aruzicka> that would be much appreciated
<aruzicka> https://github.com/Dynflow/dynflow/pull/402
<KeeperoftheKeys> looks like it works :) thnx
```

PR #406, opened 2022-01-28
Add support for halting execution plans

adamruzicka:
Up until now, Dynflow only had support for clean cancellation. Upon cancellation
a cancel event was sent to all cancellable suspended and running steps and it
was the action's responsibility to cancel itself. However, there were actions
which just weren't cancellable at all.

This commit adds support for halting execution plans. The actions are completely
unaware of this, on halt dynflow just destroys its internal state about a given
execution plan and turns it to stopped state. It does not remove running steps
from workers as we currently don't have any real means of doing so. Any pending
events will be rejected. In other words, halting an execution plan ensure it
will not run any further.

### DIFF
--- a/examples/halt.rb
+++ b/examples/halt.rb
@@ -56,7 +56,7 @@ class Wrapper < Dynflow::Action
   end
 end
 
-if $0 == __FILE__
+if $PROGRAM_NAME == __FILE__
   puts example_description
 
   ExampleHelper.world.action_logger.level = Logger::DEBUG
--- a/examples/halt.rb
+++ b/examples/halt.rb
@@ -1,0 +1,71 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative 'example_helper'
+
+example_description = <<DESC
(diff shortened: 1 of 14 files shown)


ezr-ondrej:
This sounds like a nice feature, but I do not feel comfortable enough to ellaborate on the consequences here...

ezr-ondrej:
@adamruzicka if you'd like me to give it another try, I can do a review here. I've improved my code reading skills and I've learned YOLO, so I'd be happy to merge after a proper review

adamruzicka:
I was about to merge this myself in the upcoming weeks, but I'll never say no to another pair of eyes on a pr so if you could, I'd be grateful for that

ezr-ondrej (approved):
Overall looks pretty well designed :)
There is one too many `halted` flags to my taste and I'm unsure if the clean-up is all correct, but looks like it! :)
+1 from me FWIW.

adamruzicka:
Thank you @ezr-ondrej !

PR #408, opened 2022-02-11
Migrate forgotten columns to msgpack

adamruzicka:


ezr-ondrej (approved):
Thanks @adamruzicka :+1:
Sidenote: I always wonder why it is still so difficult to review code that had just moved around :(

PR #411, opened 2022-02-28
Use ::ActiveRecord::Base.configurations.configs_for

stejskalleos:
Fixing changes introduced in https://github.com/Dynflow/dynflow/pull/410, where `::Rails.application.config.database_configuration[::Rails.env]` doesn't load data from the `DATABASE_URL` variable


adamruzicka:
Thank you @stejskalleos !

PR #414, opened 2022-05-25
load config in a way that's compatible with ActiveRecord 6.0 and 6.1

evgeni:
this makes the change from bd0b3d71c9403da553e98ed2c75bf1c568315cf9 backwards compatible


adamruzicka (approved):
Tested on my development setup and on a hijacked month-old foreman nightly. Works well on both, hopefully this will solve the issue for good

adamruzicka:
Thank you @evgeni !

PR #418, opened 2022-09-07
Limit update SQL queries to only columns that changed

adamruzicka:
For testing run dynflow with `DYNFLOW_SQL_LOG=1`


ares (approved):
Thanks @adamruzicka, merging!

PR #422, opened 2023-02-03
Fix oddities in delayed planning

adamruzicka:
Dynflow has a built in subcomponent which runs inside the orchestrator and periodically dispatches delayed execution plans scheduled for the future. Once the delayed plan is properly planned, the delay record (the thing saying "an execution plan $X should be executed at time $T") is destroyed. There are some safeguards in place to ensure a single delayed plan does not get planned multiple times. 

Issue 1:
In sidekiq-based deployments, the delayed plan dispatching subcomponent is started too early, while the rest of the orchestrator is still doing world validity checks. This can lead to a situation where the subcomponent dispatches a single delayed plan multiple times.

Issue 2:
When delayed plans get dispatched multiple times, the safeguards are not handling it properly. The safeguards essentially act as an early return in case the plan in question is already being planned, however, as soon as the early return happens, the delayed record is removed. This breaks planning of the next repetition, which relies on data from it.

This PR addresses ~issue 2~ both issues

### DIFF
--- a/lib/dynflow/director.rb
+++ b/lib/dynflow/director.rb
@@ -167,7 +167,7 @@ module Dynflow
       @logger = world.logger
       @execution_plan_managers = {}
       @rescued_steps = {}
-      @planning_plans = []
+      @planning_plans = Set.new
     end
 
     def current_execution_plan_ids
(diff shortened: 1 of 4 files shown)


ofedoren (approved):
Thanks, @adamruzicka, it's been for a while here, let's get this in.

PR #427, opened 2023-04-12
Survive minor db drops

adamruzicka:
Essentially it means wrapping all db interactions in the persistence adapter with retries. We don't really need the retries, but we need to wrap the db-specific errors with general dynflow persistence errors which the rest of the codebase expects.


ofedoren:
Thanks, @adamruzicka !

IIUC, this will help if a DB won't be available for cca 1 minute, and if more it will just kill the process now? If so, apologies for the newbie question, but looking at the related tickets, how will that help users to recover from the DB drop for e.g. 15 minutes? Is it a clear state that dynflow process is dead and needs to be started, is it that something will automatically try to restart the process since it's dead now?

Issue #431, opened 2023-10-19
warnings when booting app with sidekiq 6.5.x


evgeni:
Ohai,

we've recently upgraded Foreman to use sidekiq 6.5.x, and now when we boot the app the following warnings are logged:

```
Oct 19 06:31:08 centos8-stream-foreman-nightly.tanso.example.com foreman[1267]: 2023-10-19T06:31:08.167Z pid=1267 tid=33b WARN: `config.options[:key] = value` is deprecated, use `config[:key] = value`: ["/usr/share/gems/gems/dynflow-1.8.0/lib/dynflow/rails/configuration.rb:103:in `sidekiq_worker?'", "/usr/share/gems/gems/dynflow-1.8.0/lib/dynflow/rails/configuration.rb:155:in `block in world_config'"]
Oct 19 06:31:08 centos8-stream-foreman-nightly.tanso.example.com foreman[1268]: 2023-10-19T06:31:08.171Z pid=1268 tid=334 WARN: `config.options[:key] = value` is deprecated, use `config[:key] = value`: ["/usr/share/gems/gems/dynflow-1.8.0/lib/dynflow/rails/configuration.rb:103:in `sidekiq_worker?'", "/usr/share/gems/gems/dynflow-1.8.0/lib/dynflow/rails/configuration.rb:155:in `block in world_config'"]
Oct 19 06:31:08 centos8-stream-foreman-nightly.tanso.example.com foreman[1266]: 2023-10-19T06:31:08.174Z pid=1266 tid=33a WARN: `config.options[:key] = value` is deprecated, use `config[:key] = value`: ["/usr/share/gems/gems/dynflow-1.8.0/lib/dynflow/rails/configuration.rb:103:in `sidekiq_worker?'", "/usr/share/gems/gems/dynflow-1.8.0/lib/dynflow/rails/configuration.rb:155:in `block in world_config'"]
Oct 19 06:31:08 centos8-stream-foreman-nightly.tanso.example.com foreman[1266]: 2023-10-19T06:31:08.206Z pid=1266 tid=33a WARN: `config.options[:key] = value` is deprecated, use `config[:key] = value`: ["/usr/share/gems/gems/dynflow-1.8.0/lib/dynflow/rails/configuration.rb:103:in `sidekiq_worker?'", "/usr/share/gems/gems/dynflow-1.8.0/lib/dynflow/rails/configuration.rb:99:in `increase_db_pool_size?'"]
[...]

PR #432, opened 2023-10-19
Access sidekiq config using a non-deprecated way

adamruzicka:
like
```
WARN: `config.options[:key] = value` is deprecated, use `config[:key] = value`
```

#431


ofedoren (commented):
Thanks, @adamruzicka, LGTM.

Although, I think I found one more place to change:
 - https://github.com/Dynflow/dynflow/blob/master/lib/dynflow.rb#L28

Should we also update examples then? Such as https://github.com/Dynflow/dynflow/blob/master/examples/remote_executor.rb#L76

adamruzicka:
Ah yes, good catch

ofedoren (approved):
Thanks, @adamruzicka, :shipit:

adamruzicka:
Thank you @ofedoren !

PR #438, opened 2023-11-09
Call the on_planning_finished on cancel

adamruzicka:
This will only happen if there were any sub-plans which were not planned yet when the cancel event was processed.


ofedoren (approved):

lhellebr:
LGTM, as long as `total_count` and `output` aren't accessed concurrently.

PR #439, opened 2023-11-20
**kwargs extravaganza

adamruzicka:
good enough to make examples/future_execution and examples/orchestrate_evented work

Let's see how many tests will be broken by this

### DIFF
--- a/lib/dynflow/actors/execution_plan_cleaner.rb
+++ b/lib/dynflow/actors/execution_plan_cleaner.rb
@@ -43,7 +43,7 @@ module Dynflow
         def clean!
           plans = @world.persistence.find_old_execution_plans(Time.now.utc - @max_age)
           report(plans)
-          @world.persistence.delete_execution_plans(uuid: plans.map(&:id))
+          @world.persistence.delete_execution_plans(uuid: plans.map(&:id), **{})
         end
 
         def report(plans)
--- a/examples/future_execution.rb
+++ b/examples/future_execution.rb
@@ -14,23 +14,33 @@ end
 
 class CustomPassedObjectSerializer < ::Dynflow::Serializers::Abstract
   def serialize(arg)
-    # Serialized output can be anything that is representable as JSON: Array, Hash...
-    { :id => arg.id, :name => arg.name }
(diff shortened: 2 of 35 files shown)


ofedoren (commented):
Thanks, @adamruzicka, the changes look logical, but I don't understand why it raises
`NoMethodError: undefined method 'to_hash' for Support::DummyExample::Polling:Class`. It seems `Support::DummyExample::Polling` is `Action`, which should (?) respond `to_hash` via `recursive_to_hash` (?). Obviously something gets broken on a way to serialization, but no idea when.

ofedoren:
Seems okay-ish?.. I mean, do we need to pass `**{}`? Does it fail without this on Ruby 2.7?

adamruzicka:
I'm afraid so. In some place we explicitly need to tell ruby to not treat the previous argument as an options hash (and not call to_hash on it) :/

lib/dynflow/actors/execution_plan_cleaner.rb
```diff
@@ -43,7 +43,7 @@ def start
         def clean!
           plans = @world.persistence.find_old_execution_plans(Time.now.utc - @max_age)
           report(plans)
-          @world.persistence.delete_execution_plans(uuid: plans.map(&:id))
+          @world.persistence.delete_execution_plans(uuid: plans.map(&:id), **{})
```
ofedoren, line 46: Should we have `**{}` here? I mean, `delete_execution_plans` looks to have `def delete_execution_plans(filters, batch_size = 1000, backup_dir = nil)`. It won't hurt, but it just seems unnecessary. On the other hand, I've noticed this: https://github.com/Dynflow/dynflow/blob/master/lib/dynflow/debug/telemetry/persistence.rb#L34-L38. I guess it makes more sense to add **kwargs support there just in case.

ofedoren (commented):
Thanks, @adamruzicka, I didn't find any issues now, you did incredible work here :D

Just one this inline and I guess it's ready.

adamruzicka:
In the end we decided to not go down this path.

PR #442, opened 2024-01-25
Update rubocop to a slightly less ancient version to allow us to use ruby-2.7 syntax

adamruzicka:


adamruzicka:
@ofedoren not great, I admit, but I don't really have the determination to fix all gorillion cops that were added or changed since rubocop-0.39.0. On the brighter side, this should allow us to use the triple dot splat.

lib/dynflow.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
+
 require 'apipie-params'
 require 'algebrick'
-require 'thread'
```
ofedoren, line 4: Just to make sure, this isn't being used, is it?

ofedoren (commented):
Thanks, @adamruzicka, love the changes :) The only question though:

lib/dynflow.rb
adamruzicka: We're using threads, but it seems to be auto-required. Or to put it the other way around, we don't seem to be using anything related to threads that we'd have to explicitly require

ofedoren (approved):
Thanks, @adamruzicka, LGTM :)

PR #444, opened 2024-02-21
Suppress serialization warnings from Sidekiq

adamruzicka:
There are some changes coming in Sidekiq 7 that will need changes on our side and currently Sidekiq emits lots of warnings about it. We are aware of the issue, but until we resolve it properly, it doesn't make sense to spam all the users' logs with it.


ofedoren (approved):

PR #445, opened 2024-05-14
Fix deprecation warning

sbernhard:
This fixes the following deprecation warning:
WARN: `config.options[:key] = value` is deprecated, use `config[:key] = value`: ["/usr/share/gems/gems/dynflow-1.8.2/lib/dynflow.rb:28:in `block in process_world'", "/usr/share/gems/gems/sidekiq-6.5.12/lib/sidekiq.rb:142:in `configure_server'"]


adamruzicka (approved):
APT

adamruzicka:
Thank you @sbernhard !

PR #446, opened 2024-05-17
Add execution plan chaining

adamruzicka:
This commit enables execution plans to be chained. Assuming there is an execution plan EP1, another execution plan EP2 can be chained onto EP1. When chained, EP2 will stay in scheduled state until EP1 goes to stopped state. An execution plan can be chained onto multiple prerequisite execution plans, in which case it will be run once all the prerequisite execution plans are stopped.

It builds on mechanisms which were already present. When an execution plan is chained, it behaves in the same way as if it was scheduled for future execution. A record is created in dynflow_delayed_table and once the conditions for it to execute are right, it is dispatched by the delayed executor. Because of this, there might be small delay between when the prerequisites finishs and the chained plan is started.

TODOs:
- ? somehow show the prerequisite execution plans in the console

### DIFF
--- a/lib/dynflow/persistence_adapters/sequel.rb
+++ b/lib/dynflow/persistence_adapters/sequel.rb
@@ -39,7 +39,8 @@ module Dynflow
                     envelope:            %w(receiver_id),
                     coordinator_record:  %w(id owner_id class),
                     delayed:             %w(execution_plan_uuid start_at start_before args_serializer frozen),
-                    output_chunk:        %w(execution_plan_uuid action_id kind timestamp) }
+                    output_chunk:        %w(execution_plan_uuid action_id kind timestamp),
+                    execution_plan_dependency: %w(execution_plan_uuid blocked_by_uuid) }
 
       SERIALIZABLE_COLUMNS = { action:  %w(input output),
                                delayed: %w(serialized_args),
@@ -139,12 +140,16 @@ module Dynflow
           .all.map { |plan| execution_plan_column_map(load_data plan, table_name) }
       end
 
-      def find_past_delayed_plans(time)
+      def find_ready_delayed_plans(time)
         table_name = :delayed
         table(table_name)
-          .where(::Sequel.lit('start_at <= ? OR (start_before IS NOT NULL AND start_before <= ?)', time, time))
+          .left_join(TABLES[:execution_plan_dependency], execution_plan_uuid: :execution_plan_uuid)
+          .left_join(TABLES[:execution_plan], uuid: :blocked_by_uuid)
+          .where(::Sequel.lit('start_at IS NULL OR (start_at <= ? OR (start_before IS NOT NULL AND start_before <= ?))', time, time))
+          .where(::Sequel[{ state: nil }] | ::Sequel[{ state: 'stopped' }])
           .where(:frozen => false)
           .order_by(:start_at)
+          .select_all(TABLES[table_name])
           .all
           .map { |plan| load_data(plan, table_name) }
       end
@@ -157,6 +162,10 @@ module Dynflow
 
       def save_delayed_plan(execution_plan_id, value)
         save :delayed, { execution_plan_uuid: execution_plan_id }, value, with_data: false
+      end
+
+      def chain_execution_plan(first, second)
+        save :execution_plan_dependency, { execution_plan_uuid: second }, { execution_plan_uuid: second, blocked_by_uuid: first }, with_data: false
       end
 
       def load_step(execution_plan_id, step_id)
@@ -297,7 +306,8 @@ module Dynflow
                  envelope:            :dynflow_envelopes,
                  coordinator_record:  :dynflow_coordinator_records,
                  delayed:             :dynflow_delayed_plans,
-                 output_chunk:        :dynflow_output_chunks }
+                 output_chunk:        :dynflow_output_chunks,
+                 execution_plan_dependency: :dynflow_execution_plan_dependencies }
 
       def table(which)
         db[TABLES.fetch(which)]
--- a/lib/dynflow/persistence_adapters/sequel_migrations/025_create_execution_plan_dependencies.rb
+++ b/lib/dynflow/persistence_adapters/sequel_migrations/025_create_execution_plan_dependencies.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  up do
+    type = database_type
+    create_table(:dynflow_execution_plan_dependencies) do
+      column_properties = if type.to_s.include?('postgres')
+                            { type: :uuid }
+                          else
+                            { type: String, size: 36, fixed: true, null: false }
+                          end
+      foreign_key :execution_plan_uuid, :dynflow_execution_plans, on_delete: :cascade, **column_properties
+      foreign_key :blocked_by_uuid, :dynflow_execution_plans, on_delete: :cascade, **column_properties
+      index :blocked_by_uuid
+    end
+  end
+
+  down do
+    drop_table(:dynflow_execution_plan_dependencies)
+  end
+end
--- a/examples/execution_plan_chaining.rb
+++ b/examples/execution_plan_chaining.rb
@@ -1,0 +1,42 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require_relative 'example_helper'
+
+class DelayedAction < Dynflow::Action
+  def plan
+    plan_self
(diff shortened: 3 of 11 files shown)


ianballou:
This will be helpful to solve the issue in Katello where a composite content view auto publish can be triggered for content views that are still publishing. They are separate execution chains that need to rely on each other.

lib/dynflow/persistence_adapters/sequel.rb
```diff
@@ -139,12 +140,16 @@ def find_old_execution_plans(age)
           .all.map { |plan| execution_plan_column_map(load_data plan, table_name) }
       end
 
-      def find_past_delayed_plans(time)
+      def find_ready_delayed_plans(time)
```
ianballou: I had an issue where one composite CV publish was waiting on two children component CV publishes. However, the task would still start after the quicker child finished.

I don't know Dynflow super well, so I employed some AI tool help:

---
/begin robot

Issue: The find_ready_delayed_plans query in lib/dynflow/persistence_adapters/sequel.rb had a bug when handling execution plans with multiple dependencies. 

It would return a delayed plan as "ready" if ANY dependency was stopped, instead of waiting for ALL dependencies to stop. 

Root Cause: The original query used LEFT JOINs:

```
LEFT JOIN dependencies ON delayed.uuid = dependencies.execution_plan_uuid
LEFT JOIN execution_plans ON dependencies.blocked_by_uuid = execution_plans.uuid
WHERE (state IS NULL OR state = 'stopped')
```

With multiple dependencies (e.g., plan D depends on A and B):

If A is 'running' and B is 'stopped', the LEFT JOIN produces 2 rows
The WHERE clause filters out the row with A ('running')
But keeps the row with B ('stopped')
Result: D is returned as "ready" even though A is still running

Fix: Changed to NOT EXISTS subquery to ensure NO dependencies are in a non-stopped state:

```
WHERE NOT EXISTS (
  SELECT 1 FROM dependencies
  LEFT JOIN execution_plans ON dependencies.blocked_by_uuid = execution_plans.uuid
  WHERE dependencies.execution_plan_uuid = delayed.execution_plan_uuid
  AND execution_plans.state IS NOT NULL
  AND execution_plans.state != 'stopped'
)
```

Result: Chained execution plans now correctly wait for ALL dependencies to complete before running, as documented in the original PR description.

/end robot

---

I tested this out, and afterwards the publish did indeed wait properly for the slower child to finish.

It's possible I'm using this chaining method incorrectly in my development branch, but let me know what you think of the above.

adamruzicka: Oh yeah, this was put together rather quickly, I'll have to take a look at this again

ianballou:
I'm planning on submitting a Foreman Tasks PR for integration there. In Katello I currently am including some ForemanTasks addons for the proof of concept.

ianballou:
Is there a good way to get the input data from the scheduled chained task? So far it looks like the inputs aren't available yet so they need to be grabbed from `ForemanTasks.dynflow.world.persistence.load_delayed_plan(task.external_id).args`.

ianballou:
Is there a good way to get the input data from the scheduled chained task? So far it looks like the inputs aren't available yet so they need to be grabbed from `ForemanTasks.dynflow.world.persistence.load_delayed_plan(task.external_id).args`.

> > Is there a good way to get the input data
> 
> How would you define input data? What you get from `ForemanTasks.dynflow.world.persistence.load_delayed_plan(task.external_id).args` are the arguments that will end up being passed to `#plan`. The root action's `#input` may end up being different after the whole planning goes through.
> 
> What would be the use case?

Okay, bear with me here.

When publishing component 2+ CVs that are part of a composite with auto publish enabled, 2+ composite CV publishes will be triggered. This is bad because the latter publish will error out on the Lock being taken. Plus, since the parent composite publish will wait on its publishing component children, only 1 composite CV publish is needed.

So, my implementation in Katello needs to look up the chained & scheduled composite CV publish task in order to skip auto publishing if there is already a task waiting. And thus I need to access the arguments that were passed to plan (which I was calling the input, but I suppose that is different from the actual task input). From the arguments I can find the content view ID to ensure that it is the composite waiting to be published.

ianballou:
> So, my implementation in Katello needs to look up the chained & scheduled composite CV publish task in order to skip auto publishing if there is already a task waiting.

I am rethinking this. If a child publishes while the parent is publishing, we would want a new CCV version to be published. That would mean chaining the new composite content view publish on the scheduled/running one. However, when I tried this, there were Lock errors. Could it work such that the lock checking is deferred to when the scheduled task actually runs?

Katello has a workaround for this via the Event queue, and it polls on lock failures. When it was implemented 6 years ago though chaining in Dynflow was wished for :) https://github.com/Katello/katello/pull/8188

ianballou:
@adamruzicka I've tested this a bunch now with https://github.com/Katello/katello/pull/11540.

It seems to be working really well. The scheduled task waits on its children and cancels itself if one of them fails.

Do you need me to perform a code review as well? Or can we have someone with an outside opinion come do that?

lib/dynflow/persistence_adapters/sequel_migrations/025_create_execution_plan_dependencies.rb
```diff
@@ -0,0 +1,21 @@
...
+    type = database_type
+    create_table(:dynflow_execution_plan_dependencies) do
+      column_properties = if type.to_s.include?('postgres')
+                            { type: :uuid }
+                          else
+                            { type: String, size: 36, fixed: true, null: false }
+                          end
+      foreign_key :execution_plan_uuid, :dynflow_execution_plans, on_delete: :cascade, **column_properties
+      foreign_key :blocked_by_uuid, :dynflow_execution_plans, on_delete: :cascade, **column_properties
+      index :blocked_by_uuid
```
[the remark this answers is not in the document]
adamruzicka: Added

lib/dynflow/delayed_plan.rb
```diff
@@ -31,6 +31,12 @@ def timeout
       error("Execution plan could not be started before set time (#{@start_before})", 'timeout')
     end
 
+    def failed_dependencies(uuids)
+      bullets = uuids.map { |u| "- #{u}" }.join("\n")
+      msg = "Execution plan could not be started because some of its preqrequisite execution plans failed:\n#{bullets}"
```
sjha4: Typo in `prerequisite`

web/views/show.erb
```diff
@@ -43,6 +43,32 @@
   <%= h(@plan.ended_at) %>
 </p>
 
+<% if @plan.state == :scheduled && @plan.delay_record %>
```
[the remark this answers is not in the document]
sjha4: Don't know if it's presentable on the UI but the chaining being displayed only while chained task is scehduled and waiting seems like losing information after the job runs on the dynflow UI..Might be good to persist the UI chain info for debugging reasons?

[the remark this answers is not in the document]
ianballou: Do you mean like, for example, change from "Waiting for execution plans:" to "Waited for execution plans:" once the task actually runs? And start showing the old chained plans, of course.

web/views/show.erb
```diff
@@ -43,18 +43,16 @@
...
-      <% dependencies.each do |dep_id| %>
-        <li><a href="<%= url("/#{dep_id}") %>"><%= h(dep_id) %></a></li>
-      <% end %>
-      </ul>
-    </p>
-  <% end %>
+<% dependencies = @plan.world.persistence.find_execution_plan_dependencies(@plan.id) %>
+<% if dependencies.any? %>
+  <p>
+    <b>Depends on execution plans:</b>
```
ianballou: @sjha4 it looks like what you asked for, a permanent dependency history, is here now :)

sjha4: Awesome..It was hard to test the dependencies on the UI without this.. :)

ianballou:
The only thing I haven't been able to test yet is multiple dependencies showing up. I'd like to get that going first, and then we should be good to merge here. Testing has been solid so far.

lib/dynflow/persistence_adapters/sequel.rb
```diff
@@ -175,6 +195,10 @@ def save_delayed_plan(execution_plan_id, value)
         save :delayed, { execution_plan_uuid: execution_plan_id }, value, with_data: false
       end
 
+      def chain_execution_plan(first, second)
```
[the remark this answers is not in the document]
adamruzicka: I took a different approach, but it should be fixed.

PR #447, opened 2025-01-30
Edit how failed_count and cancelled_count is calculated

adamlazik1:
This PR proposes to mark every sub plan which finishes as not successful
after the parent execution plan receives the cancel directive as
cancelled. This is to be consistent with how a task is evaluated as
cancelled in the foreman project.

This PR has an additional effect which makes the parent execution plan
always finish with warning if at least one sub plan is cancelled as a
response to the `cancel` event. Until now, it could either finish with
success if all subplans either finished with success or were not yet
queued for execution or it could finish with warning if already pending
sub plan was cancelled and finished with the `error` result.  In
summary, this additional change unifies the behavior of a cancelled
execution plan if there is anything left to cancel.

### DIFF
--- a/lib/dynflow/action/v2/with_sub_plans.rb
+++ b/lib/dynflow/action/v2/with_sub_plans.rb
@@ -102,7 +102,7 @@ module Dynflow::Action::V2
 
     def increase_counts(planned, failed)
       output[:planned_count] += planned + failed
-      output[:failed_count]  = output.fetch(:failed_count, 0) + failed
+      output[:failed_count] = output.fetch(:failed_count, 0) + failed
       output[:pending_count] = output.fetch(:pending_count, 0) + planned
       output[:success_count] ||= 0
     end
@@ -129,12 +129,20 @@ module Dynflow::Action::V2
     end
 
     def recalculate_counts
-      total   = total_count
-      failed  = sub_plans_count('state' => %w(paused stopped), 'result' => %w(error warning))
+      total = total_count
+      if output[:cancelled_timestamp]
+        cancelled_scheduled_plans = sub_plans_count_after(output[:cancelled_timestamp], { 'state' => %w(paused stopped), 'result' => %w(error warning) })
+        cancelled_unscheduled_plans = total_count - output[:planned_count]
+        cancelled = cancelled_unscheduled_plans + cancelled_scheduled_plans
+      else
+        cancelled = cancelled_scheduled_plans = 0
+      end
+      failed = sub_plans_count('state' => %w(paused stopped), 'result' => %w(error warning)) - cancelled_scheduled_plans
       success = sub_plans_count('state' => 'stopped', 'result' => 'success')
-      output.update(:pending_count => total - failed - success,
-                    :failed_count  => failed - output.fetch(:resumed_count, 0),
-                    :success_count => success)
+      output.update(:pending_count => total - failed - success - cancelled_scheduled_plans,
+        :failed_count => failed - output.fetch(:resumed_count, 0),
+        :success_count => success,
+        :cancelled_count => cancelled)
     end
 
     def counts_set?
@@ -142,7 +150,7 @@ module Dynflow::Action::V2
     end
 
     def check_for_errors!
-      raise SubtaskFailedException.new("A sub task failed") if output[:failed_count] > 0
+      raise SubtaskFailedException.new("A sub task failed") if output[:failed_count] + output[:cancelled_count] > 0
     end
 
     # Helper for creating sub plans
@@ -173,6 +181,7 @@ module Dynflow::Action::V2
     def cancel!(force = false)
       # Count the not-yet-planned tasks as cancelled
       output[:cancelled_count] = total_count - output[:planned_count]
+      output[:cancelled_timestamp] ||= Time.now.utc.iso8601 # time in UTC for comparison with UTC times in the database
       on_planning_finished if output[:cancelled_count].positive?
       # Pass the cancel event to running sub plans if they can be cancelled
       sub_plans(:state => 'running').each { |sub_plan| sub_plan.cancel(force) if sub_plan.cancellable? }
@@ -198,7 +207,9 @@ module Dynflow::Action::V2
     end
 
     def remaining_count
-      total_count - output[:cancelled_count] - output[:planned_count]
+      return 0 if output[:cancelled_timestamp]
+
+      total_count - output[:planned_count]
     end
 
     private
@@ -215,6 +226,10 @@ module Dynflow::Action::V2
 
     def sub_plans_count(filter = {})
       world.persistence.find_execution_plan_counts(filters: sub_plan_filter.merge(filter))
+    end
+
+    def sub_plans_count_after(timestamp, filter = {})
+      world.persistence.find_execution_plan_counts_after(timestamp, { filters: sub_plan_filter.merge(filter) })
     end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/dynflow/persistence_adapters/abstract.rb
```diff
@@ -46,6 +46,10 @@ def find_execution_plan_counts(options = {})
         filter(:execution_plan, options[:filters]).count
       end
 
+      def find_execution_plan_counts_after(options = {}, timestamp)
+        filter(:execution_plan, options[:filters]).filter(::Sequel.lit('ended_at >= ?', timestamp)).count
```
adamruzicka, line 50: The abstract adapter shouldn't rely on any implementation-specific details (ie. `Sequel.lit`). If that can't be done, it is better to just raise `NotImplementedError`

lib/dynflow/action/v2/with_sub_plans.rb
```diff
@@ -173,6 +175,7 @@ def concurrency_limit_capacity
     def cancel!(force = false)
       # Count the not-yet-planned tasks as cancelled
       output[:cancelled_count] = total_count - output[:planned_count]
+      output[:cancelled_timestamp] = DateTime.now.new_offset(0).to_s # time in UTC for comparison with UTC times in the database
```
adamruzicka, line 178: I'd rather be explicit with the format here to save ourselves from any suprises
```suggestion
      output[:cancelled_timestamp] = DateTime.now.new_offset(0).iso8601 # time in UTC for comparison with UTC times in the database
```

adamruzicka: Also this reads easier (or at least to me)
```suggestion
      output[:cancelled_timestamp] = Time.now.utc.iso8601 # time in UTC for comparison with UTC times in the database
```

adamruzicka, line 178: In theory, `cancel!` can be called multiple times, in which case it should keep the oldest timestamp

adamruzicka (commented):
The general approach looks good, left a couple of comments. Also rubocop is red and I'm not sure what the tests will say

adamlazik1:
Thanks, I applied the suggestions. I noticed in the dynflow console that for whatever reason cancelled RunHostsJob shows result success even though the counts are correct. On master it shows result = error if a job is cancelled. Is this a problem?

lib/dynflow/persistence.rb
```diff
@@ -73,6 +73,10 @@ def find_execution_plan_counts(options)
       adapter.find_execution_plan_counts(options)
     end
 
+    def find_execution_plan_counts_after(options, timestamp)
```
ofedoren, line 76: Shouldn't be the parameters switched? It seems to be inconsistent with the usages like in https://github.com/Dynflow/dynflow/pull/447/files#diff-c58910824d8fc847b39d994b727a0dd9958e732ad1551e0e9b6fa62cccc6ede8R227. I don't even know how that would not fail...

```suggestion
    def find_execution_plan_counts_after(timestamp, options)
```

adamlazik1: Thanks for catching that. I first had options in the first place but I had to switch it because optional arguments should be last. I forgot about this one place.

adamlazik1:
Alright, unless I am missing something I have fixed the issue with different result state. Right now the only change should be in the cancelled and failed counts of a cancelled job.

adamlazik1:
Ugh, I see a problem when cancelling more jobs and I wonder what should be the correct implementation. The table below displays result of RunHostsJob in different scenarios. The job on 1 host is cancelled before finishing and the job on two hosts is run with concurrency level 1 and is cancelled after it successfully completes on the first host.

| n. of hosts / branch | master  | PR
------------------------------------------------------
| 1 | warning | warning
| 2 | success | warning

Under what circumstances should the result of RunHostsJob be warning and success when cancelling a job?

adamlazik1:
> Ugh, I see a problem when cancelling job invocation on more than 1 hosts and I wonder what should be the correct implementation. The table below displays result of RunHostsJob in different scenarios. The job invocation on 1 host is cancelled before finishing and the job invocation on two hosts is run with concurrency level 1 and is cancelled after the command run successfully completes on the first host.
> n. of hosts / branch 	master 	PR
> 1 	warning 	warning
> 2 	success 	warning
> 
> Under what circumstances should the result of RunHostsJob be warning and success when cancelling a job invocation?

After discussion with @adamruzicka we concluded that the behavior on master could be considered an inconsistency and thus the side effect of this PR in its current form is actually a plus, so I am keeping the behavior for now. I updated the commit message accordingly.

I made some additional errors in the last version and I hope I fixed them in the current one, but I need to test more scenarios so I will flip this to draft for now.

Also, I need to transfer a value (`cancelled_unqueued_sub_plans_count`) between different methods and I don't know what the correct way of doing that here is so I used the `output` hash for now. Suggestions are welcome.

adamruzicka:
> Also, I need to transfer a value (cancelled_unqueued_sub_plans_count) between different methods and I don't know what the correct way of doing that here is so I used the output hash for now. Suggestions are welcome.

Can't you calculate that from the other counts that you have?

adamlazik1:
I am not sure how I would do that. Those unqueued subplans aren't even in the database as far as I understand, no?

adamlazik1:
> I am not sure how I would do that. Those unqueued subplans aren't even in the database as far as I understand, no?

Looking at the code I could calculate it from total_count and planned_count, but I don't know where total_count even comes from as it appears to be (?) a virtual method. Can I count on it always returning the same value during one job invocation run? Same question goes for planned_count. From the code and output when running execution plans it seems that it always stays the same, but is there any scenario in which this value could change inbetween the cancel event occurence and the next call of recalculate_counts?

adamlazik1:
On a separate note: I retested the current version and it appears to be working as expected and the errors I had seem have been resolved. I am flipping this back to Ready for review, even though it is likely that more changes are pending based on the above comments.

adamlazik1:
> > I am not sure how I would do that. Those unqueued subplans aren't even in the database as far as I understand, no?
> 
> Looking at the code I could calculate it from total_count and planned_count, but I don't know where total_count even comes from as it appears to be (?) a virtual method. Can I count on it always returning the same value during one job invocation run? Same question goes for planned_count. From the code and output when running execution plans it seems that it always stays the same, but is there any scenario in which this value could change inbetween the cancel event occurence and the next call of recalculate_counts?

After another discussion it turns out relying on dynamic calculation of unscheduled sub plans from total_count and planned_count should be save, thus I no longer need to transfer this value inbetween functions. PR updated and ready for review.

I hope I was able to edit the `remaining_count` method correctly. In the previous implementation `cancelled_count` could either be 0 or `total_count - planned_count` at the time of cancel event, which means that at the time of cancellation this method should return 0.

adamruzicka:
Yeah, these do happen from time to time.

adamruzicka:
Thank you @adamlazik1 !

lib/dynflow/action/v2/with_sub_plans.rb
```diff
@@ -129,20 +129,28 @@ def run_progress
     end
 
     def recalculate_counts
-      total   = total_count
-      failed  = sub_plans_count('state' => %w(paused stopped), 'result' => %w(error warning))
+      total = total_count
+      if output[:cancelled_timestamp]
+        cancelled_scheduled_plans = sub_plans_count_after(output[:cancelled_timestamp], { 'state' => %w(paused stopped), 'result' => %w(error warning) })
```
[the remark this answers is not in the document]
adamlazik1: If it failed after marked as cancelled then I would assume it should be ok to count it as cancelled.

PR #468, opened 2026-04-18
A very primitive Actor implementation

jturel:
I was able to pull something basic together pretty quickly. It works with my related Katello change https://github.com/jturel/katello/pull/1/changes Of course, it's lacking in the finer details.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -169,7 +169,7 @@ Metrics/CyclomaticComplexity:
 # Offense count: 135
 # Configuration parameters: CountComments, CountAsOne, ExcludedMethods.
 Metrics/MethodLength:
-  Max: 47
+  Max: 48
 
 # Offense count: 4
 # Configuration parameters: CountComments, CountAsOne.
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -169,7 +169,7 @@ Metrics/CyclomaticComplexity:
 # Offense count: 135
 # Configuration parameters: CountComments, CountAsOne, ExcludedMethods.
 Metrics/MethodLength:
-  Max: 48
+  Max: 44
 
 # Offense count: 4
 # Configuration parameters: CountComments, CountAsOne.
--- a/lib/dynflow/coordinator.rb
+++ b/lib/dynflow/coordinator.rb
@@ -188,7 +188,7 @@ module Dynflow
       end
     end
 
-    class ActorLock < LockByWorld
+    class SingletonActorLock < LockByWorld
       def initialize(world, actor_name)
         super(world)
         @data[:id] = self.class.lock_id(actor_name)
--- a/lib/dynflow/coordinator.rb
+++ b/lib/dynflow/coordinator.rb
@@ -188,6 +188,17 @@ module Dynflow
       end
     end
 
+    class AgentLock < LockByWorld
+      def initialize(world, agent_name)
+        super(world)
+        @data[:id] = self.class.lock_id(agent_name)
+      end
+
+      def self.lock_id(agent_name)
+        "agent:#{agent_name}"
+      end
+    end
+
     class DelayedExecutorLock < LockByWorld
       def initialize(world)
         super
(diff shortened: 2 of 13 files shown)


lib/dynflow/world.rb
```diff
@@ -70,6 +70,36 @@ def initialize(config)
...
       end
       post_initialization
+      @agents = {}
+    end
+
+    def find_agent(name)
+      @agents[name]
+    end
+
+    def register_agent(name, value:, observers: [])
```
adamruzicka: Rather than being able to register agents on the fly, I was thinking about going more declarative route. World configuration would have an `extra_agents` option (not set on the name), which would accept a hash name-class pairs.

When used with foreman, foreman-tasks would provide an option through which plugins could register their agents and foreman-tasks would pass the hash with all the agents to dynflow when it actually initializes it.

lib/dynflow/world.rb
```diff
@@ -78,6 +108,7 @@ def initialize(config)
     def post_initialization
       @delayed_executor ||= try_spawn(:delayed_executor, Coordinator::DelayedExecutorLock)
       @execution_plan_cleaner ||= try_spawn(:execution_plan_cleaner, Coordinator::ExecutionPlanCleanerLock)
+      # TODO: is an agent executor needed?
```
adamruzicka: No, I don't think so.

lib/dynflow/coordinator.rb
```diff
@@ -188,6 +188,17 @@ def self.inherited(klass)
...
+    class AgentLock < LockByWorld
+      def initialize(world, agent_name)
+        super(world)
+        @data[:id] = self.class.lock_id(agent_name)
+      end
+
+      def self.lock_id(agent_name)
+        "agent:#{agent_name}"
+      end
+    end
```
adamruzicka: This would mean two things:
- the agent would be globally unique
- the agent would only be spawned during world initialization

I don't see any mechanism that would cause other world to try and start the agent if the world that currently holds the lock went away. We could either have the worlds themselves retry or always spawn the agents in all the worlds, but let them compete for the lock.

.rubocop_todo.yml
```diff
@@ -169,7 +169,7 @@ Metrics/CyclomaticComplexity:
 # Offense count: 135
 # Configuration parameters: CountComments, CountAsOne, ExcludedMethods.
 Metrics/MethodLength:
-  Max: 47
+  Max: 48
```
adamruzicka: Eh, sure, I don't feel strongly about this either way.

adamruzicka (commented):
I finally had some time to read this and the concurrent ruby docs again and I must admit I was a bit rusty on it. It seems an [actor](https://ruby-concurrency.github.io/concurrent-ruby/1.1.5/Concurrent/Actor.html) was the thing I had in mind. Considering agents are more about state, an actor would most likely be a better fit. Sorry about that.

Apart from the katello queue, I'd eventually like to migrate this [mqtt re-sending part of sp-rex-ssh](https://github.com/theforeman/smart_proxy_remote_execution_ssh/blob/master/lib/smart_proxy_remote_execution_ssh/mqtt/dispatcher.rb) to this rather just spawning actors ad-hoc.

.rubocop_todo.yml
jturel: I wasn't sure if rubocop would gate running of tests so I quickly did this. No intention of keeping it

lib/dynflow/coordinator.rb
jturel: I was thinking more toward the latter but only through the lens of "how many places does the applicability host queue need to be cleared from?" Other use cases may not want global uniqueness at all.

jturel:
Moving to draft as this is really just an experiment at the moment.

lib/dynflow/coordinator.rb
jturel: > I don't see any mechanism that would cause other world to try and start the agent if the world that currently holds the lock went away

Were you asking this simply from the point of safety or is there a general case where a world may go away and not be replaced by another that would restart the actor?

adamruzicka: Mostly from the point of safety.

General cases could be found once in multi-executor deployments, but those are pretty rare and even then you'd have to try hard to trigger it.

PR #472, opened 2026-08-19
Fixes #471 - Tolerate absent duplicate indexes

### LINKED ISSUE #471: Duplicate index migration fails when an index is already absent
## Problem

`020_drop_duplicate_indices` unconditionally drops three indexes that duplicate primary-key constraints. If one of those indexes is already absent, `dynflow:migrate` aborts with `PG::UndefinedObject` instead of accepting that index as already removed.

This prevents Foreman from completing its database migration. The failure was reported in [Foreman Redmine #31736](https://projects.theforeman.org/issues/31736) with Dynflow 1.4.7, and migration 020 remains functionally unchanged in Dynflow 2.0.1.

## Expected behavior

Migration 020 should treat an already-absent duplicate index as the desired state and continue removing any remaining duplicate indexes.

## Proposed fix

Check the indexes exposed by Sequel before dropping each one, and add a regression test for a schema where one duplicate index is already missing before migration 020 runs.

## AI usage disclosure

This issue was investigated and prepared with the assistance of Codex 5.6 Sol High. The findings were reviewed before submitting.

jakduch:
## Summary

Fixes #471.

Related: [Foreman Redmine #31736](https://projects.theforeman.org/issues/31736).

Migration 020 unconditionally dropped indexes that duplicate primary-key constraints. When one of those indexes was already absent, the migration aborted instead of accepting the schema as already being in the desired state.

Check the indexes reported by Sequel before dropping each duplicate. This keeps the migration portable across the supported PostgreSQL and SQLite adapters and allows it to continue removing the remaining duplicate indexes.

Add a regression test that migrates through version 19, removes one duplicate index, and verifies that migration 20 succeeds and removes the other two.

## Testing

- `test/persistence_migrations_test.rb`: 1 run, 6 assertions, 0 failures, 0 errors

## AI usage disclosure

Per the [community discussion on AI policy](https://community.theforeman.org/t/could-foreman-benefit-from-an-ai-usage-policy/44638), the issue was investigated and the changes, tests, and PR wording were prepared with the assistance of Codex 5.6 Sol High. The resulting changes were reviewed before submitting. The commit also includes an `Assisted-By` trailer.


lib/dynflow/persistence_adapters/sequel_migrations/020_drop_duplicate_indices.rb
```diff
@@ -2,16 +2,22 @@
 
 Sequel.migration do
   up do
-    alter_table(:dynflow_actions) do
-      drop_index [:execution_plan_uuid, :id]
+    if indexes(:dynflow_actions).key?(:dynflow_actions_execution_plan_uuid_id_index)
```
adamruzicka, line 5: For the actual dropping, we rely on sequel to derive the index name from the table and the columns. Could we reuse that function rather than using the exact index name?

adamruzicka (commented):
471? Now this is some stuff of the legend.

One comment inline

lib/dynflow/persistence_adapters/sequel_migrations/020_drop_duplicate_indices.rb
jakduch: Yes, good point. `default_index_name(table, Array(columns)).to_sym` will keep the existence check aligned with the name Sequel uses in `drop_index`. I’ll update it.

